PR #84, opened 2017-04-16
added Pcons page but didn't do the settings yet

Hour-of-the-Owl:
also /tp gh doesn't work so I removed that from chat commands


reduf:
I think we should just add `/tp gh` to the commands.

PR #86, opened 2017-04-16
asdf

DarkManic:


docs/settings.md
```diff
@@ -0,0 +1,16 @@
+---
```
HasKha: removed settings_folder.md and added this settings.md with less content

PR #126, opened 2017-07-17
Added a presearing check on loaded zone.

stheno:
This changes the zone list based on character loaded and repopulates the list between character switching.


HasKha:
I like the idea, but I see some issues in the implementation. I haven't tested the code, but looking at the PR I have the following comments:

- Global variables are definitely to be avoided. Use fields of the object if it makes sense, or enclose them in the unnamed namespace. 
- IsPresearing implies a boolean return value, yet it's an integer, which is later mostly used as a boolean.
- Presearing test should be done once every frame, not during initialization, and not in helped functions. 
- We could use the newly introduced mapinfo (https://github.com/GregLando113/GWCA/commit/b3e8831) which in theory is cleaner, but game updates won't happen so it's not really a problem.
- I noticed you remove the usual buttons while in pre-searing. I like that, but why not add the pre-searing towns instead? There are few enough of them to fit the current layout. That might even completely remove the need for the drop-down menu. 
- Please do not add random namespaces. Add functionality to the class or to the unnamed namespace in the .cpp file.

HasKha:
Implemented by @reduf in https://github.com/HasKha/GWToolboxpp/commit/719d0bb7b9f4c3a430a82ad8012926922072272b, sorry @stheno, closing this.

stheno:
ok, np

PR #138, opened 2017-11-16
Fix missing "Go" Button in TravelWindow.

Sil3ntStorm:
Favorites didn't have Go button and couldn't be used.


HasKha:
Thanks, I forgot about those when updating to the new imgui release.

PR #145, opened 2017-12-29
Team hero builds

thx4nothing:
Added a new Window for creating Team Builds for heroes that can be loaded with one button (or hotkey).
Players won't have to select every hero separately anymore.

### DIFF
--- a/GWToolbox/GWToolbox/Modules/ChatCommands.cpp
+++ b/GWToolbox/GWToolbox/Modules/ChatCommands.cpp
@@ -49,7 +49,7 @@ void ChatCommands::DrawHelp() {
 	ImGui::Bullet(); ImGui::Text("'/dialog <id>' sends a dialog.");
 	ImGui::Bullet(); ImGui::Text("'/flag [all|<number>]' to flag a hero in the minimap (same a using the buttons by the minimap).");
 	ImGui::Bullet(); ImGui::Text("'/hide <name>' closes the window or widget titled <name>.");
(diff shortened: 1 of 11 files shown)


HasKha:
Thanks for the PR. Code looks good overall. A few comments:

* If multiple skills are allowed, then user should be able to set their usage with a single command, e.g. /useskill 135 or /useskill 1 3 5 (imho both should be valid). 
* Multiple skills usage requires prioritization, that's something that became clear when I had that feature in old toolbox. Essentially /useskill 351 should prioritize skill 3, 5, then 1, if more than one is off-cooldown. Without this, people would for example /useskill 123 on emo and die because the software would keep using burning speed instead of ER. This probably requires a change in the datastructure, 
* I like the idea of the herobuild feature, and I'm sure many users will use it, however I don't like adding a default main window item for a feature that many others will not use. Thinking about it, we probably want to add an option to show items in the main window, similar to how we can completely enable/disable them. I'll see if I can implement that, otherwise note that I might default it to disabled on release (user can still enable it). 
* I want to play with the hero builds more but should be ok. Will take a more detailed look later.

HasKha:
> Thinking about it, we probably want to add an option to show items in the main window, similar to how we can completely enable/disable them

Okay I did that, so don't worry about it.

HasKha:
working on this

HasKha:
@thx4nothing did you have a reason for leaving the build name there, or did you just not change it?

I'm thinking on completely removing build names, and replacing that with the hero name. Would you be against that? If so, why?

thx4nothing:
I didn't see a reason to delete it. I think it could be useful for some people. I like to know which roll a hero has.
For example, I use it like this:
htb

Issue #120, opened 2017-06-02
Check in which instances Toolbox is already running on startup


AnonymousII:
Whenever tb is started while more than one instance of gw is running it always displays a combobox where you have to choose in which gw instance tb should be injected. Even in situations where this makes absolutely no sense, e.g. when you got two gw instances and tb is already running in one of them - starting tb twice in the same gw instance doesnt make any sense, so what else should the user choose in this popup?

So it would be nice to check on tb startup in which gw instances tb is already injected and filter them out of the list - if theres only one gw instance without tb left just start it there and dont ask where to start. Should be really easy to implement - e.g. just list the loaded dlls for each gw instance and check for tb dlls.

It would also be nice to have a button "Inject into all gw instances" so you dont have to do it for every instance manually.

HasKha:
I tend to prefer to keep the launcher as simple as possible, to avoid both antivirus false positive detection and various injection bugs, but this is a very good point and we should do it if possible.

AnonymousII:
the easiest solution is probably just making a list of gw window handles and/or names and put it in some temporary file - I would be surprised if this behaviour triggers an antivirus detection :)

reduf:
Fixed in 7d24e64
Not that it will still show the char selector even if you can only inject one process, because it is not necessary obvious in which client you want to inject the toolbox.

PR #162, opened 2018-02-15
enable multilaunch

AnonymousII:
Real implementation of #120 - Ability to inject into multiple / all GW instances

New Features:
1) Inject in multiple / all running GW instances
2) easy check / uncheck all
3) changed combo box into a checked list box
4) keeping reduf's feature to only list GW instances without TB / GWCA injected

Screenshot of adapted gui:
zwischenablage01

Code complexity stays about the same, still easy to understand and maintain, shouldnt do any more av triggers than initial implementation

### DIFF
--- a/CSLauncher/CSLauncher.cs
+++ b/CSLauncher/CSLauncher.cs
@@ -28,7 +28,7 @@ namespace CSLauncher {
         };
 
 
-        static Process proctoinject = null;
+        //static Process proctoinject = null;
  
         [STAThread]
         static void Main(string[] args) {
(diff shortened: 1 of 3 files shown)


HasKha:
Thanks for the PR. I am a bit conflicted about this, because while the feature is nice to have, I don't know if this kind of workflow (injecting toolbox on multiple clients) is used by many users, and some might prefer the current launcher. 

I think we can definitely make both options available, but I'll also see if we can add this functionality in a way that is not too different from the current behavior, for example by adding a "Inject All" button, or an advanced mode that opens this window.

AnonymousII:
The easiest way to combine is probably to select the topmost entry by default and keep the rest unchecked. This way the default behaviour is exactly the same, by clicking launch it will simply inject into the first client :) And people that want to inject tb in many / all instances can just tick up the others or use Select All button

HasKha:
thanks again for the PR. I implemented this but leaving the default behavior untouched. I mostly used your code, so thanks again. Closing.

PR #169, opened 2018-03-02
Maintain Field of View

### LINKED ISSUE #168: Camera FOV maintain through zoning
Camera FOV is not maintained through zoning. You have to use the command every time you zone. I'm unsure how to run a camera command upon loading into a zone, so I'm looking for someone to point me in the right direction. I was thinking of attaching it to a packet but I can't tell what one to attach it to.

ritlew:
I added "Maintain FOV" as an option to the settings. I was tired of having to maintain it myself every time I opened the game or rezoned. Let me know if this isn't the legitimate way to do this.

closes #168

### DIFF
--- a/GWToolbox/GWToolbox/Modules/ChatCommands.cpp
+++ b/GWToolbox/GWToolbox/Modules/ChatCommands.cpp
@@ -23,6 +23,7 @@
 #include "GWToolbox.h"
 #include <Keys.h>
 #include <logger.h>
+#include <Modules\GameSettings.h>
 #include <Windows\MainWindow.h>
 #include <Windows\SettingsWindow.h>
 #include <Windows\TravelWindow.h>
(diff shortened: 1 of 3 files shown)


HasKha:
We tried to implement this in the past, with a similar implementation (if (GetFov() != fov) SetFov(fov) ) and it caused subtle bugs and crashes on map change. 
It is possible that since then the CameraMgr and overall GWCA robustness improved, but I would still like to make sure this change is stable. Have you (extensively) tested it?

ritlew:
I had it crash once while I was developing it. However it hasn't crashed since I had this working code. When it crashed I was doing something a little different by changing values in GWCA,

ritlew:
Well I've been zoning for awhile with no crashes, but I suppose we need it tested on other machines than mine.

ritlew:
Closing and resubmitting on different branch

PR #184, opened 2018-03-27
Custom colors for minimap

MikeJerred:
Added color for winnowing spirit alongside eoe & qz.
Changed soul tormentor counter to only show when in DoA.
Added custom agent colors for the minimap that are saved in the INI.


GWToolbox/GWToolbox/Widgets/Minimap/AgentRenderer.cpp
```diff
@@ -248,16 +296,44 @@ void AgentRenderer::Render(IDirect3DDevice9* device) {
 			&& (npcs[agent->PlayerNumber].NpcFlags & 0x10000) > 0) continue;
 		if (target == agent) continue; // will draw target at the end
 
+		bool hasCustomColor = false;
```
HasKha: I don't understand this block and the next one. Why do you collect custom agents and render them separately? Modifying GetColor(Agent*) should be enough.

GWToolbox/GWToolbox/Widgets/Minimap/AgentRenderer.cpp
```diff
@@ -324,6 +400,11 @@ Color AgentRenderer::GetColor(GW::Agent* agent) const {
 		}
 	}
 
+	auto it = std::find_if(colors_custom.cbegin(), colors_custom.cend(), [=](const std::pair<DWORD, Color>& pair) { return pair.first == agent->PlayerNumber; });
+	if (it != colors_custom.cend()) {
+		return it->second;
+	}
+
```
HasKha: I'm worried about performance here, if the user has a large number of custom colors, searching through them for every agent render can be expensive, probably unordered_map<playernumber, color> would be better.

HasKha: Also I think that custom colors should still darken a bit when below 90%, but feel free to disagree with me as you have more experience with that. We could add an extra color as a darkening factor to apply to all others while <90%.

GWToolbox/GWToolbox/Widgets/Minimap/AgentRenderer.cpp
MikeJerred: You don't want to have a group of enemies and can't see the colored fiend because it is hidden under the others. The list of custom colors is meant to be an ordered list, so what is at the top gets rendered first. I should probably add controls to move the color up/down as well as delete though

GWToolbox/GWToolbox/Widgets/Minimap/AgentRenderer.cpp
MikeJerred: Yeah I agree with this. We can remove the < 90% enemy color and replace it with a darken mask that gets applied to everything?

HasKha: sure, that works

MikeJerred: I'm not worried 💃 
I would prefer to wait and see if this is actually a performance problem before spending the time to change it: as the colors are ordered by what should render first we can't simply change the container to a map.

MikeJerred:
I shall add a bit of complexity in the UI because I think it's important to be able to at least delete custom colors, so hold off on merging this for now at least.

GWToolbox/GWToolbox/Widgets/Minimap/AgentRenderer.cpp
MikeJerred: I'm not worried 💃 but yes we can copy the vector into a map before iterating through the agents.

MaddoW: But please let there be a checkbox to enable/disable the darkening feature. Not everyone had different colors for >90% / <90% enemies.

GWToolbox/GWToolbox/Widgets/Minimap/AgentRenderer.cpp
MaddoW: This is a great Idea, to bring different colored enemies on top, since I had that coloring feature for some enemies, I always thought about that, but was to lazy.

GWToolbox/GWToolbox/Widgets/Minimap/AgentRenderer.cpp
HasKha: You can just set the darkening color to (0,0,0,0) and it'll do nothing.

MaddoW: Perfect

HasKha:
Thanks. Do you think it's worth it having a separate ini file? Similar to Markers.ini

MikeJerred:
It would be a bit more organised. Probably makes more sense to do that for hotkeys, for me at least I have about 20 hotkeys but only 3 colored enemies.

HasKha:
Yea I guess we can do that too. Initially all settings were in one GWToolbox.ini but I've been splitting them. I moved teambuilds out of the main ini in 2.6, we can do the same for hotkeys.

HasKha:
I merged, but looking at the code I still have a doubt: why collect custom colored agents, *sort them*, and render them?

I understand we want to render the colored agents after all the default ones, but what is the sort for? And as an alternative, could we loop and render non-colored first, and then loop again and render colored after? (without building the array of custom_agents).

HasKha:
Also about the interface, I think it would make more sense to match the UI style of the custom markers, rather than that of the hotkeys.

MikeJerred:
The sorting is because I want them to be sorted :P if I have a large group of soul tormentors, I need to be able to see if one fiend has spawned in the middle of them, so I need fiends to be rendered on top of souls for example.

I mean you can change the custom_agents vector to a list, or build it as a map and loop through the colors_custom vector instead of sorting if you are so worried about performance - but I really think it is effort spent on a non-issue.

PR #210, opened 2018-05-20
website update

Hour-of-the-Owl:
I haven't written anything about custom agents or hero builds yet - will get on that when I have time.


HasKha:
There is a conflict on the main page table. It was changed to a native HTML table. 

You probably did not pull the changes before committing and making your PR. Try to do that in the future.

For now we can probably solve it here. What did you change in the table?

PR #211, opened 2018-05-22
Added /pings chat command

tjubutsi:
I'm not too sure about some stuff but at least it's a few things already done.


tjubutsi:
So toggle between normal and 10% visible pings? Something like that?

HasKha:
yea, that could be done as well

tjubutsi:
Is it useful to keep this? I don't think I know enough to create the whole settings thing myself.

HasKha:
yes, I'll merge and work on it when I have time. I prioritized the other PR despite being later as we've been actually thinking on that for a while longer.

HasKha:
Ended up implementing it with a different solution, check the above commit for details. Closing.

PR #215, opened 2018-05-30
Maddo w master

reduf:
Made it simpler


reduf:
Not what I meant to do.

PR #218, opened 2018-06-07
pls fix

tjubutsi:


reduf:
I don't get the "## Version 2.12```". Is that an error ?

tjubutsi:
Yes

PR #288, opened 2019-01-21
GWCA is no longer private

kyleshrader:
Removing the setup work required for using GWCA-Public. This is no longer needed.

I'm sure there are other things that can be cleaned up, but my understanding of the project is pretty shallow, so i'll stick to just updating the readme.


HasKha:
Thanks!

PR #290, opened 2019-01-27
Ctrl Shift Click item to open wiki

kyleshrader:
Adding option to ctrl shift click an item to open the wiki page for that item.

This is just an idea I had when i was going through a bunch of items in my storage, cleaning things out. Feel free to discuss the feature and try it out for yourself.

I added the option right next to ctrl+click items to move to storage, and used functionality similar to the scwiki chat command to perform opening the wiki in a browser.

### DIFF
--- a/GWToolbox/GWToolbox/Modules/GameSettings.cpp
+++ b/GWToolbox/GWToolbox/Modules/GameSettings.cpp
@@ -1,6 +1,7 @@
 #include "GameSettings.h"
 
 #include <ctime>
+#include <ShellApi.h>
 
 #include <GWCA\Managers\GameThreadMgr.h>
 #include <GWCA\Managers\MapMgr.h>
@@ -12,6 +13,7 @@
 #include <GWCA\Managers\FriendListMgr.h>
 #include <GWCA\Managers\Render.h>
 #include <GWCA\Managers\CameraMgr.h>
+#include <GWCA\Managers\UIMgr.h>
 
 #include <GWCA\Context\GameContext.h>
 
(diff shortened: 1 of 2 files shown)


GWToolbox/GWToolbox/Modules/GameSettings.cpp
```diff
@@ -868,10 +876,26 @@ bool GameSettings::WndProc(UINT Message, WPARAM wParam, LPARAM lParam) {
 }
 
 void GameSettings::ItemClickCallback(uint32_t type, uint32_t slot, GW::Bag *bag) {
-	if (!GameSettings::Instance().move_item_on_ctrl_click) return;
+	if (!GameSettings::Instance().move_item_on_ctrl_click && !GameSettings::Instance().wiki_item_on_ctrl_shift_click) return;
     if (!ImGui::IsKeyDown(VK_CONTROL)) return;
 	if (type != 7) return;
```
3vcloud: This would only trigger wiki if item was type 7

GWToolbox/GWToolbox/Modules/GameSettings.cpp
```diff
@@ -868,10 +876,26 @@ bool GameSettings::WndProc(UINT Message, WPARAM wParam, LPARAM lParam) {
...
 
+	// Ctrl + Shift + Click will open the gww with the item name as a search term.
+	if (ImGui::IsKeyDown(VK_SHIFT)) {
+		GW::Item *item = GW::Items::GetItemBySlot(bag, slot + 1);
+		if (!item || !item->NameString) return;
+
+		std::wstring itemName;
+		GW::UI::AsyncDecodeStr(item->NameString, &itemName);
```
3vcloud: is itemName guaranteed to be decoded by the time you create the wiki link below?

HasKha:
Thanks for the PR, and thanks for the comments, I agree and they should be addressed.

I really don't think this feature should be hardcoded to ctrl shift, because I don't expect many users to use this and it is very possible to have this enabled, and triggering it by mistake in an explorable (holding ctrl+shift is common for finding targets), causing gw to lose focus. Any issue like that would bring even more people to disable this.

It's not a bad feature, but I believe it's very niche, and we may have to develop a better way to implement things like this (key + item click), allowing users to select the key combination, or at least the behavior on give common key combinations.

kyleshrader:
Very good point. I think that might be ctrl and alt to show targets, but whatever's the case, it could still be disruptive. Disabling by default isn't really helpful either, because then it's just an unused feature. In most UI designs, a context menu on right click is used.

reduf:
I would want to see a need for this feature, before locking an other hotkey.

Right click on the item -> "Open on the Wiki" would be the best, but we can't really do that. Well maybe...

PR #369, opened 2019-08-02
Update DialogsWindow.cpp

Hour-of-the-Owl:
everyone always uses hex!


BearsMan:
Misty, I don't think it matters because hex and dec are similar for the dialogs ill give u an example: when I put the DoA dialogs in I used Dec not Hex

reduf:
I accept it, because in info window we give the dialog value in hex. Make sense to have it by default.

PR #383, opened 2019-11-25
Coindrop hotkey is no longer needed

BearsMan:
https://gyazo.com/48bc05083a69a69b516aba144a3ca954 - removed from line 369 since Coindrop hotkey no longer works


reduf:
Indeed, thank you!

BearsMan:
Livia im gonna re do the change because commit wont go through im sorting it out

Issue #311, opened 2019-03-10
Mission Completion and Title status info?


dfarley1:
I recently started digging around in the GWToolbox and GWCA code to see if I could implement something like a panel that would show what missions a character has done for Protector/Guardian.  I'm not sure where I'm going with it but I wanted to see if I could implement some widget like #119 requested, automatically accept Ziashen quests that the character needed, TP to relevant outputs, stuff like that.

Anyways, I'm unable to find any context (GameContext, WorldContext, Titles) that show character mission progress that's already been found.  Do you guys have any hints or tips that I could explore?

3vcloud:
https://github.com/GregLando113/GWCA/blob/8a0e5a522e417dd3e7d39d97a4157d41dbc49f4a/Source/GWCA/Context/WorldContext.h#L74 might be a good idea to check whether a player has access to a map, and https://github.com/GregLando113/GWCA/blob/8a0e5a522e417dd3e7d39d97a4157d41dbc49f4a/Source/GWCA/Context/WorldContext.h#L61 might show you completed quests.

Thats where I'd start, but haven't really looked into those arrays.

reduf:
You can look if you already found a map by using the following array.
https://github.com/GregLando113/GWCA/blob/master/Source/GWCA/Context/WorldContext.h#L74

e.g.
```cpp
bool GetIsMapUnlocked(uint32_t map_id)
{
    Array<uint32_t> unlocked_map = GameContext::instance()->world->unlocked_map;
    uint32_t index = map_id / 32;
    if (index >= unlocked_map.size())
        return false;
    uint32_t shift = map_id % 32;
    uint32_t flag = 1 << shift;
    return (unlocked_map[index] & flag) != 0;
}
```

dfarley1:
Thanks so much for the leads guys, I noticed unlocked_map before but had no idea how to start deciphering it except trying to track it as I played the game and did some missions!  I'll check when I get home but I wonder if those values show the mission completion (masters, bonus, etc.) or Hard Mode.

reduf:
Doing something similar as an example of how to use GWCA. If it get popular enough, might move to GWToolbox.

PR #392, opened 2019-12-23
Add Mission tracker window

dfarley1:
This PR adds a Missions window that shows completion status for all NM and HM missions and dungeons. It also supports highlighting ZM/ZB missions for "today" as well as any previous quests in the player's quest log.

I had this idea a while ago but couldn't implement it until @reduf pointed out his GWCA example in #311. I copied the original logic/constants from that example but have refactored it several times while developing this.


reduf:
Hi, so the main question for us is always if something add values without increasing the friction too much and we will discuss that.

In the meantime I tested it a bit and it's quite nice, but some small comments **not** regarding the code:
- The icon in the main window doesn't fit with the others.
- I would like the have a button to switch between the normal mode and the hard mode directly in the window. There is a bunch of cities (e.g. Guild Hall) where you don't have the ingame option to switch.
- Eye of the North and Dungeons are bugged. If you expand the respective "headers", you can switch to the hard mode view but not back to the normal mode. Also, both doesn't show the completed missions.

HiImFlo:
If this gets implemented (and I see merit in doing so) it would be good to have a csv export so people can manage multiple accounts/characters in external tables.

dfarley1:
> Hi, so the main question for us is always if something add values without increasing the friction too much and we will discuss that.
> 
> In the meantime I tested it a bit and it's quite nice, but some small comments **not** regarding the code:

Sure thing. I didn't know if you guys had a style guide or anything like that and I basically tried to copy/paste as much from the Pcons and Hotkeys windows as I could so let me know if you see room for improvement!

> * The icon in the main window doesn't fit with the others.

I couldn't readily find anything that fit, where'd you get the current set of icons from? I'm trash at drawing so please don't make me make one!

> * I would like the have a button to switch between the normal mode and the hard mode directly in the window. There is a bunch of cities (e.g. Guild Hall) where you don't have the ingame option to switch.

Good idea, I'll add that in the top right or something. Is there already support for making that switch in Toolbox/GWCA? I haven't looked into any of the packet stuff but any pointers would be much appreciated!

I also had the idea for a button that teleports you to the current daily mission outpost but didn't add it yet. Also if there's a way to auto-accept quests while in an outpost could we make a "Accept daily and go to that mission" button?

> * Eye of the North and Dungeons are bugged. If you expand the respective "headers", you can switch to the hard mode view but not back to the normal mode. Also, both doesn't show the completed missions.

I encountered that bug while testing and I thought I fixed it for them all, but I guess I only did for the campaigns. Whoops!  I'm getting more used to Python than C++, for shame.

That's also something on my todo list, I lifted the code for checking mission status from @reduf but his example didn't do it for EotN or dungeons so I didn't know if it worked the same.  I'll investigate and get that working too.

dfarley1:
> If this gets implemented (and I see merit in doing so) it would be good to have a csv export so people can manage multiple accounts/characters in external tables.

My original idea was an app or something that would give notifications if there was a daily you didn't have completed on one of your characters so after this made it into GWToolbox I was going to start thinking about how to hook it up to an external server for exactly that kind of account management. CSV seems like a decent first step, shouldn't be too hard to make a primitive version.

justinbeatz:
You could implement sqlite for the management, btw when this is gonna be implemented?

HiImFlo:
Hey, I still think this would be a great addition. Currently doing some character development and ofc I can manually manage stuff in excel, but this would be very convenient.

dfarley1:
I haven't opened up GW in a while and I don't know when I'm going to again, although I do regret that I didn't get this finished before I fell off. The biggest issue right now is that I've more-or-less permanently moved to Linux on my desktop and something tells me I'm not gonna be able to get VS and Toolbox running under Wine. GW is old enough that I might be able to get it working in a VM, I might try that at some point soon, but don't hold your breath :(

3vcloud:
I was going to sort this one for you (toolbox codebase has changed a lot since you did this PR). Regardin nix, one of the toolbox devs does his jazz in Linux, and there are guys in the discord that have got it working well on their nix machines

3vcloud:
This is in dev on debug, but dungeons aren't working properly and using the icons like it is isn't very user friendly - once these things have been sorted we can make it available to release, but closing this PR now that we have the code in

PR #411, opened 2020-04-29
Add Refrains to bond monitor

PostCrafter:
Since [Heroic Refrain](https://wiki.guildwars.com/wiki/Heroic_Refrain) isn't actually a bond (maintained enchantment/buff), it is not possible to see if other players, henchmen or allies are affected by it. Therefore this feature is only useful when playing by yourself with heroes - and this is honestly the main use case for this.

Additionally I've added an option to not render effects (only Heroic Refrain for now) if they were applied with an attribute lower than the current. This will prove useful when having to first stack up Heroic Refrain but already starting to apply it to heroes before being maxed out.
Unfortunately this only works when seeing the actual effect, so it's not possible to add the functionality to bonds you maintain on other players.


HasKha:
That's pretty cool! 

As a suggestion, what about showing the lower-attribute skill differently rather than having an option to hide it? For example, it could be shaded differently, or have a border.

PostCrafter:
Great suggestion!
I've thought about it but haven't put any time into finding out how that would work, will definitely do.

PostCrafter:
Is there some decent way of greyscaling loaded textures in code or should I just add grey versions of the images?

3vcloud:
you could always overlay the image with a semi trans colour

PostCrafter:
I've replaced the checkbox with a color selection (in case anyone wants to overlay it differently).
This looks quite nice now.
image

3vcloud:
Nice, have you taken a look at the damage monitor to figure out how tall these icons need to be?

PostCrafter:
I've only added the 64x64 image to the bond monitor and didn't change anything about how they're rendered.

PostCrafter:
As of now I don't care about a lot of other skills, partially because I didn't play a lot of paragon yet.
If someone wants to have other skills (enchantments etc) included in this, please let me know which.

PostCrafter:
I've experimented adding every Enchantment, Shout, Weaponspell, Chant and Echo to the bond monitor.
I did this by adding a single sprite including every skill icon and mapping each SkillID to the appropriate frame on the image (not sure if this is the correct approach here, but it certainly sounded like less effort than adding every single skill icon as a resource).
The image is almost 4MB in size, so I don't know how desireable it is. It would allow us to display arbitrary skill icons throughout the application though (custom hero panels or showing skills in the build window comes to mind).

3vcloud:
This isn't a good idea, no - it'll make the dll massive and as it is it will only use a few kb of that whole file.

PostCrafter:
I absolutely agree, which is why I haven't included it in this PR.
Is there some reasonable way to render the images from Guild Wars directly?
If there isn't, does anyone have a strong opinion on skills to add in order to move forward with this PR?

3vcloud:
You can find a skill's file id from its struct, which pretty much maps to the hash in the DAT file. From there you could somehow pull the skill image out, but it was too much effort and time for me to investigate how

3vcloud:
Can you add a section to enable/disable individual bonds to the settings for the widget please? want to add this to bond monitor but refrains off unless the player turns it on

3vcloud:
Added in https://github.com/HasKha/GWToolboxpp/compare/2949b297f490...d4c675f1cd13 - having used it theres no need for the toggle for refrains

PostCrafter:
Okay, great 👍

bquenin:
Just out of curiosity, when you say:

> Since Heroic Refrain isn't actually a bond (maintained enchantment/buff), it is not possible to see if other players, henchmen or allies are affected by it.

Is it because the GW client/protocol doesn't provide this information, or is it not surfaced by the Toolbox dll?
Just asking because it would actually be very useful to check who's affected by the shout even with players.

DubbleClick:
The client never receives the information.

bquenin:
@DubbleClick thanks for the clarification

PR #412, opened 2020-04-29
Additional Range Indicators

PostCrafter:
This adds a few additional range indicators:
- Shadowstep aggro range: 1010 (aggro range) around your shadow step marker, to help with not breaking the stance into aggro
- Rezz aggro range: 1010 (aggro range) around your targeted dead ally, to help with rezzing them outside aggro
- Chain aggro range: 700 range around targeted foes, to help seeing when it is safe to be attacked

### DIFF
--- a/GWToolboxdll/Widgets/Minimap/AgentRenderer.h
+++ b/GWToolboxdll/Widgets/Minimap/AgentRenderer.h
@@ -24,8 +24,8 @@ public:
 	uint32_t auto_target_id = 0;
 
 private:
-	static const size_t shape_size = 5;
-	enum Shape_e { Tear, Circle, Quad, BigCircle, EmptyCircle };
+	static const size_t shape_size = 4;
+	enum Shape_e { Tear, Circle, Quad, BigCircle };
 	enum Color_Modifier {
 		None, // rgb 0,0,0
 		Dark, // user defined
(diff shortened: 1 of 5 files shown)


PostCrafter:
I'm sure there's improvements to be made regarding the rendering, so please leave some feedback, as this is my first time doing anything like this.

HasKha:
I like shadowstep and res aggro ranges, but having a constant range indicator around your target could add visual noise while not being particularly useful to a lot of players. Maybe we can default it to disabled (transparent). 

I can look over the rendering but it looks okay at a first glance (and it's been a while since I touched that code :p )

PostCrafter:
Sure, I'll split up the ranges, so I can actually assign different colors and default it to transparent for chain aggro.

PostCrafter:
I've been using this feature for almost 2 weeks now and like it quite a bit.
Before actually merging this we might want to have a look at the default values, as most people are probably not interested in chain aggro range.

Hour-of-the-Owl:
I'm mildly triggered by the colours going ARGB instead of RGBA... but more concerned that my HoS line is replaced by a weird cross that always points east. Apart from this I've encountered no problems.

3vcloud:
Would've merged this in yesterday but theres still a lot of conflicts and I didn't have the time

BearsMan:
@PostCrafter u may wanna update these :)

3vcloud:
@PostCrafter I've implemented these with a few modifications in https://github.com/HasKha/GWToolboxpp/commits/737907aa53d54305617b5ba4100a539c8b1c7d47 
 - chain aggro range is disabled by defualt, the other 2 are enabled by default

Issue #239, opened 2018-07-12
Hero flag formations.


iaankrynauw:
Spec:
- Add another button row container.
- Keep all flagging buttons consistent. (styling, double-click cancel, hover etc)
- Be able to easily add new formations. (Within code and not persisted in a file)
- Validate that heroes are placed in a valid pathing map.

First attempt will implement. 
HA style Circle
`x^2 + y^2 = r^2`
Upside down V:
`y =  -|-x| + k`
 GvG style line formations.
`y = k+a` where ` x = -2,+2`
`y = k` where ` x = -1, +1`
`y = k-a` where ` x = -3, 0, +3`

(The math is just me thinking out loud)

HasKha:
sounds good, linking to #61 for reference

quetzalsly:
This would be an amazing feature for running DoA HM with 7 heroes. It's really tedious to keep manually splitting. The best solution would be a simple keybind, where gwtoolbox monitors for a keypress like numpad0 or something and the heroes split when you press it.

iaankrynauw:
Got quite close had a prototype up, but I never deployed the code. Wasn't sure if it was clean enough and tested.

I'll find it and see if I can extract some for someone else to pickup.

HiImFlo:
Since I assume the hero team should have some sort of reference point it should be oriented towards, I think this should be
a) the current target
b) the closest hostile unit if there is no current target

3vcloud:
Flagging interrupts actions so some thought needs to be put behind how to identify when you're in combat, and what to do in that situation

iaankrynauw:
My original idea was simply to have a button such as hero flag 1. In which you click on the map and a random formation is formed around where you clicked. That could be `circle` or 2 - 3 - 2 `line` etc.

quetzalsly:
Just a random thought, but the ST hero should be behind everyone I think? he is the one that the whole team depends on to live.

FrankGriffioen:
Basicly just had the same idea, it would be great to add a button that places all hero's in a formation (by placing multiple flags at once), Optionally the formation should be able to be edited through a menu option, to suit your preferred team setup. (xy coorditnates in relation to the formation flag your placing?)

bquenin:
It looks like there's a couple commits regarding this issue, does it mean it's available already?

3vcloud:
Added in 4.1

Issue #61, opened 2017-03-27
Hero flags improvements


HasKha:
Bug fixes:
- Fix flags when managing more than 7 heroes (can happen in urgoz/deep)

There are several possible improvements in the hero flagging system:
- Improve symbols in the minimap: show the number of the hero and have a different symbol for All
- Change the mouse cursor while placing a hero flag
- Implement formations. Example: press a button (or drop-down?) which then sets all hero flags according to pre-determined rules (and show the new formation while placing them, maybe in a different color).
- Be able to place flags in the 3d world. This requires some reversing work to find either the "click on flag" or "click on world" functions.

Hour-of-the-Owl:
is it out of the question to make henchies flaggable? They are very commonly used in HA

HasKha:
you can flag them with the 'all' flag but not individually. That's not something we can add.

daanvandenhoek:
I dont know if this helps with point 5 but found this in an old au3 clickpoint file. It uses the clicked position for cords:
```
Local $iCoordAddress = ScanForPtr($hGWProcess, "D8658485FFD9C0", -4)

Local $dCoordStruct = DllStructCreate("float ClickX;float ClickY")
Local $iStructSize = DllStructGetSize($dCoordStruct)
Local $iStructPtr = DllStructGetPtr($dCoordStruct)

DllCall($hKernel32, "int", "ReadProcessMemory", "int", $hGWProcess, "int", $iCoordAddress, "ptr", $iStructPtr, "int", $iStructSize, "int*", 0)
		$fClickX1 = DllStructGetData($dCoordStruct, "ClickX")
		$fClickY1 = DllStructGetData($dCoordStruct, "ClickY")
```

3vcloud:
All of the above sorted in 4.3 with exception of hero number on minimap

PR #413, opened 2020-04-30
Add hero flagging hotkey

PostCrafter:
Adds functionality to flag heroes around the player, which makes it possible to hotkey hero formations.
Orientation is based on currently selected target or player rotation if no target is selected.
Flagging an already flagged hero instead unflags that hero.

There's plenty room for improvement (like an actual Formation Manager), but this is a decent first step.

### DIFF
--- a/GWToolbox/GWToolbox/Windows/Hotkeys.cpp
+++ b/GWToolbox/GWToolbox/Windows/Hotkeys.cpp
@@ -1,0 +1,815 @@
+#include "stdafx.h"
+#include "Hotkeys.h"
+
+#include <GWCA\Constants\Constants.h>
+#include <GWCA\GameContainers\Array.h>
+#include <GWCA\GameContainers\GamePos.h>
+
+#include <GWCA\GameEntities\Agent.h>
+#include <GWCA\GameEntities\Hero.h>
+#include <GWCA\GameEntities\Skill.h>
+
+#include <GWCA\Managers\ChatMgr.h>
+#include <GWCA\Managers\ItemMgr.h>
+#include <GWCA\Managers\AgentMgr.h>
+#include <GWCA\Managers\EffectMgr.h>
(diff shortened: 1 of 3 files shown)


HiImFlo:
This has been wished for for a couple of years and can be a feature with a lot of impact on casual PvE. Any implementation of this is great, thanks a lot! I am sure many people will really like this if it is added, since it removes one of the really tedious parts of playing with heroes.

There have been at least 2 issues regarding this, idk if you have read those but good ideas are in there: #61 and #239 

Imo the dream would be that people could use both minimap as well as a flag in 3D (with changed mouse cursor), flag the hero formation to any point they like and have the heros be oriented towards the current target like you said.

PostCrafter:
I have looked at those issues and took some inspiration (like flagging towards current target) already.

I'm currently not sure how to properly interact between mulitple modules such as pressing a hotkey/UI button to trigger a minimap formation flag on next click. Any pointers to that would be appreciated. Once that issue is out of the way I could look into adding a separate module for formation flagging.

3vcloud:
Take a look at the `/flag` chat command

3vcloud:
Sorry that was a stupid reply. You could make the var that the minimap uses a public variable and go from there, but tbh I think that you should try and find the actual in-game hook for flagging heros and use that, then we could have the minimap use it aswell instead of its own internal variable to see if the user is currently flagging. You'd be able to flag both in-game and on the minimap that way.

reduf:
It's great that you do PR, I respect that a lot, but I would recommend to do it on @3vcloud repo. I'm merely maintaining the project and maybe planning to merge @3vcloud in our repo, so a single branch exist.

3vcloud:
Added from your repo via https://github.com/HasKha/GWToolboxpp/commits/7af99d65374c8aaffd46750914b5d4abf7dedccb

PR #415, opened 2020-05-05
Add colors to distance and health monitor

### LINKED ISSUE #178: colour-changing distance and health monitors
Have the distance and health monitors appear in a different colour depending on their percent value. Distance would have one colour each for adjacent, nearby, in-the-area, aggro range (not sure this one's necessary actually? you can usually [but not always] tell if you're in aggro...), cast range, spirit range, party range. Health would be 0%, 1 to 50%, 51% to 89%, and 90% to 100%.

PostCrafter:
- Add option to change color depending on distance
- Add option to change color depending on thresholds (defaults are 50% and 90% [disabled])

Fixes #178

### DIFF
--- a/GWToolbox/GWToolbox/Widgets/DistanceWidget.cpp
+++ b/GWToolbox/GWToolbox/Widgets/DistanceWidget.cpp
@@ -80,7 +80,7 @@ void DistanceWidget::Draw(IDirect3DDevice9* pDevice) {
 			} else if (dist <= GW::Constants::Range::Compass) {
 				color = color_compass;
 			} else {
-				color = Colors::RGB(255, 255, 255);
+				color = color_widget;
 			}
 
 			Color background = Colors::RGB(0, 0, 0);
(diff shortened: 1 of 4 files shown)


PostCrafter:
Sure, I'll have a look into that later on.

PostCrafter:
Thanks for the review, I'll have a look at this tomorrow.

PostCrafter:
We currently have two options to allow changing the color of these widgets.
Foreground color:
color-foreground
Background color:
color-background
It's easy enough to allow changing either of them for the health thresholds because the setup is rather sophisticated (in fact it's done already). Adding the same possibilities for the distance monitor means duplicating all of the color selects or adding complexity to the widget.

I suggest deciding between the following options:
- making a decision for either the foreground or background coloring option above (limits the users customization possibilities)
- adding a single contrast color per widget and a toggle option to use it as foreground/background color

3vcloud:
Foreground only looks loads better

3vcloud:
Merged into dev 6f55a33eb18fc8e6e720c8e32b7105ce6da7c5d5 , post merge changes in e40cf86d4d92eb936ab1d828a4f22652bdff1810

PR #455, opened 2020-06-16
Removed SetWindowTitle()'s side-effect when enabled=false

zhro:
Changed SetWindowTitle() to no longer modify the window title of the Guild Wars window when enabled=false. Previously it would force the window title to "Guild Wars" which would override any modification made by other programs.

Recommending that this function do nothing when the Toolbox option is disabled. This is the expected behavior.


3vcloud:
Already fixed in https://github.com/HasKha/GWToolboxpp/commit/61a750985033366e727c3a12dea6d925de82b041

PR #470, opened 2020-07-08
Hotkey on toolkit start

zhro:
This is something I added for debugging purposes but I feel might be useful as a general feature. When adding new commands, it's convenient to have those commands execute at startup.


reduf:
I don't really see the use case for that. Do you have example other than during development?

zhro:
I don't. In retrospect, I should have polled users on Discord before submitted the PR to see if it other people would find it to be useful.

PR #472, opened 2020-07-12
InfoWindow does not display the model_id of item agents

zhro:
The InfoWindow was only displaying the model_id of living agents. This patch allows the model_id of item agents to be displayed as well. I wasn't able to find a source for gadget agents so they will display 0 as before.


3vcloud:
Merged into dev https://github.com/HasKha/GWToolboxpp/commits/848f31e8a3da26180a8eccc6dbef663c024336ad

PR #474, opened 2020-07-24
Add thickness slider for minimap ranges

DubbleClick:
Range circles were really really hard to see on a 4k monitor with just 1px width so I added the option to draw a couple more circles to make them more visible.


3vcloud:
Morning mate, wouldn't it be better to find a way of drawing primitives to fill the circle thickness rather than doing a while set of vertices per pixel thickness?

DubbleClick:
As a triangle list, you mean?

DubbleClick:
Okay I've tried it as both a trianglelist and using ID3DXLine with a width but the former looks terrible and the latter must've had weird issues with the transformation. I rewrote the CreateCircle method to directly generate multiple lines, but I'm not sure how else you'd prefer to have it.

PR #475, opened 2020-07-24
Fix "/pid ..." option

necropola:
* Do not ask for character name when "/pid ..." was passed
* Allow auto update to occur before injecting into specified process


reduf:
Thanks, merged in dev, so Github doesn't pick it up.

PR #476, opened 2020-07-25
option to color the background where no terrain is & circular map

DubbleClick:
the location for the map background colour might be a bit messed up (since it's in pmaprenderer now but drawn in minimap.cpp) but it's where I would've expected it in the settings

suggestions welcome

as for the circular settings that's just preparation for until I figure out how to actually only draw the circle. it still looks neat now with the background though

### DIFF
--- a/GWToolboxdll/Widgets/Minimap/PmapRenderer.cpp
+++ b/GWToolboxdll/Widgets/Minimap/PmapRenderer.cpp
@@ -56,7 +56,7 @@ void PmapRenderer::Initialize(IDirect3DDevice9* device) {
 
     // get the number of trapezoids, need it to allocate the vertex buffer
     trapez_count_ = 0;
-    for (auto &map : path_map) {
+    for (const GW::PathingMap& map : path_map) {
         trapez_count_ += map.trapezoid_count;
     }
     if (trapez_count_ == 0) return;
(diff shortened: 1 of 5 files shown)


GWToolboxdll/Widgets/Minimap/PmapRenderer.cpp
```diff
@@ -114,9 +121,8 @@ void PmapRenderer::Initialize(IDirect3DDevice9* device) {
     type = D3DPT_TRIANGLELIST;
 
     // populate vertex buffer
-    for (int k = 0; k < (shadow_show ? 2 : 1); ++k) {
-        for (size_t i = 0; i < path_map.size(); ++i) {
-            GW::PathingMap pmap = path_map[i];
+    for (auto k = 0; k < (shadow_show ? 2 : 1); ++k) {
+        for (auto pmap : path_map) {
```
reduf, line 125: Leave those as types, please.

GWToolboxdll/Widgets/Minimap/PmapRenderer.cpp
```diff
@@ -45,12 +52,12 @@ void PmapRenderer::Initialize(IDirect3DDevice9* device) {
...
         return; // no map loaded yet, so don't render anything
     }
-    bool shadow_show = ((color_mapshadow & IM_COL32_A_MASK) > 0);
+    const bool shadow_show = ((color_mapshadow & IM_COL32_A_MASK) > 0);
 
     // get the number of trapezoids, need it to allocate the vertex buffer
     trapez_count_ = 0;
-    for (size_t i = 0; i < path_map.size(); ++i) {
-        trapez_count_ += path_map[i].trapezoid_count;
+    for (auto &map : path_map) {
```
reduf, line 59: Leave the type please.

GWToolboxdll/Widgets/Minimap/PmapRenderer.cpp
DubbleClick: changed it to const GWPathingMap& for performance and your readability

PR #479, opened 2020-07-30
resharper

DubbleClick:
let resharper work this down according to the defined clang rules and this is the result


3vcloud:
I've picked the functional changes from these commits that you've done, but to avoid too many messy conflicts and a sea of code line changes in one go, please only do automatic formatting on files we're editing for a relevant bugfix/feature

PR #481, opened 2020-07-31
Partydamage height

DubbleClick:
it annoyed me to no end that the height of the party damage numbers didn't line up with the middle of the healthbar, so this fixes that


DubbleClick:
this is technically still not correct since there is a distance between health bars but after implementing it, but it's better than what was

PR #483, opened 2020-08-02
new hotkey: chain ee/recall for allies

DubbleClick:


DubbleClick:
This feature in practice (ee'd manually here, but what it would allow you to do is do this consistently): https://clips.twitch.tv/ViscousRenownedLionMoreCowbell

PR #484, opened 2020-08-04
add skillbar widget

DubbleClick:


DubbleClick:
god damnit the tracked and conflicting project files kill me

GWToolboxdll/Widgets/SkillbarWidget.cpp
```diff
@@ -0,0 +1,216 @@
...
+            m_skills[it].cooldown = skill_cooldown_to_string(std::chrono::milliseconds{skillbar->skills[it].GetRecharge()});
+
+            auto const effect_duration = get_longest_effect_duration(static_cast<GW::Constants::SkillID>(skillbar->skills[it].skill_id));
+            m_skills[it].color = UptimeToColor(effect_duration);
+        }
+    };
+    static clock_t last_update = clock();
+    if (TIMER_DIFF(last_update) >= 200) {
+        update();
+        last_update = clock();
```
HasKha, line 115: I don't like this. A 200ms delay is a ton, and the cost is negligible. Just update every frame.

GWToolboxdll/Widgets/SkillbarWidget.cpp
```diff
@@ -0,0 +1,216 @@
...
+{
+    if (!visible)
+        return;
+    const auto update = [this]() {
+        auto const *skillbar = GW::SkillbarMgr::GetPlayerSkillbar();
+        if (skillbar == nullptr)
+            return;
+
+        for (auto it = 0u; it < 8; it++) {
+            m_skills[it].cooldown = skill_cooldown_to_string(std::chrono::milliseconds{skillbar->skills[it].GetRecharge()});
```
HasKha, line 106: This seems overkill, along with all the functions at the top. Just format in-line with imgui. Reference: https://github.com/HasKha/GWToolboxpp/blob/master/GWToolboxdll/Widgets/TimerWidget.cpp#L99

GWToolboxdll/Widgets/SkillbarWidget.cpp
```diff
@@ -0,0 +1,216 @@
...
+
+            auto const effect_duration = get_longest_effect_duration(static_cast<GW::Constants::SkillID>(skillbar->skills[it].skill_id));
```
HasKha, line 108: Should we count the number of effects of a given skill, and/or both the min and max duration? Or list them all? I thought the main appeal for this kind of monitor was to see the enchantment stacking, for example for solo doa.

GWToolboxdll/Widgets/SkillbarWidget.cpp
```diff
@@ -0,0 +1,216 @@
...
+            m_skills[it].color = UptimeToColor(effect_duration);
+        }
+    };
+    static clock_t last_update = clock();
+    if (TIMER_DIFF(last_update) >= 200) {
+        update();
+        last_update = clock();
+    }
+
+    constexpr auto wnd_flags = ImGuiWindowFlags_NoTitleBar | ImGuiWindowFlags_NoResize | ImGuiWindowFlags_NoScrollbar | ImGuiWindowFlags_NoCollapse | ImGuiWindowFlags_NoInputs;
```
HasKha, line 118: Use `GetWinFlags()` to behave consistently with other Widgets, and show resize controls on all of them when needed. 
Reference: https://github.com/HasKha/GWToolboxpp/blob/master/GWToolboxdll/Widgets/TimerWidget.cpp#L98

Edit: call every frame, of course

GWToolboxdll/Widgets/SkillbarWidget.cpp
```diff
@@ -0,0 +1,216 @@
...
+    ImGui::PushStyleVar(ImGuiStyleVar_ItemSpacing, {0, 0});
+    ImGui::PushStyleVar(ImGuiStyleVar_FrameBorderSize, 1);
+    ImGui::PushStyleColor(ImGuiCol_Text, color_text);
+    ImGui::PushStyleColor(ImGuiCol_Border, color_border);
+    {
+        for (auto const &skill : m_skills) {
+            ImGui::PushID(&skill);
+            ImGui::PushStyleColor(ImGuiCol_Button, skill.color);
+            {
+                ImGui::Button(skill.cooldown.c_str(), {static_cast<float>(m_width), static_cast<float>(m_height)});
```
HasKha, line 131: Why use a button instead of a simple ImGui::Text ?

GWToolboxdll/main.cpp
```diff
@@ -17,7 +17,7 @@ DWORD WINAPI init(HMODULE hModule) noexcept {
         DWORD **found = (DWORD **)GW::Scanner::Find(
             "\xA3\x00\x00\x00\x00\xFF\x75\x0C\xC7\x05", "x????xxxxx", +1);
         if (!(found && *found)) {
-            MessageBoxA(0, "We can't determine if the character is ingame.\nContact the developpers.", "GWToolbox++ - Clientside Error Detected", 0);
+            MessageBoxA(0, "We can't determine if the character is ingame.\nContact the developers.", "GWToolbox++ - Clientside Error Detected", 0);
```
HasKha, line 20: 👍

GWToolboxdll/Widgets/SkillbarWidget.h
```diff
@@ -0,0 +1,48 @@
...
+
+    std::array<Skill, 8> m_skills;
+    Color color_text{};
+    Color color_border{};
+    Color color_long{};
+    Color color_medium{};
+    Color color_short{};
+
+    int m_width = 50;
+    int m_height = 50;
```
HasKha, line 45: maybe change the name to show that those are the skill size, not of the overall widget.

GWToolboxdll/Widgets/SkillbarWidget.cpp
HasKha: I see that this is more complex, and I like the use of modern c++, but I don't like this back-and-forth conversion from char buffers and std::string. ImGui wants char buffers, maybe keep a char buffer for each skill and snprintf in there?

GWToolboxdll/Widgets/SkillbarWidget.cpp
DubbleClick: Hmm I didn't think of solo doa as a use case. Effect stacking is rather negligible in the content I play so this is best left to someone else to implement.

GWToolboxdll/Widgets/SkillbarWidget.cpp
DubbleClick: done

GWToolboxdll/Widgets/SkillbarWidget.h
DubbleClick: done

GWToolboxdll/Widgets/SkillbarWidget.cpp
DubbleClick: For the background colour of individual skills.

quoting ocornut: "There's no concept or feature of background color for text."

GWToolboxdll/Widgets/SkillbarWidget.cpp
DubbleClick: yeah, these functions were originally used as project-wide helper methods, but since they're not used anywhere else in TB, might as well replace them by operating straight on the char buffer

DubbleClick: changed to change the member directly

DubbleClick:
added a check to exclude hexes from being coloured so this should be ready for merge now

GWToolboxdll/Widgets/SkillbarWidget.cpp
DubbleClick: added the effect time display, but it's messy. Might go over the code again

GWToolboxdll/Widgets/SkillbarWidget.cpp
```diff
@@ -26,19 +26,18 @@ namespace
 
     auto ms_to_string_secf(std::array<char, 16> &arr, std::chrono::milliseconds const cd, const char *fmt = "%.1f") -> void
```
3vcloud, line 27: don't use `std::array` and just use a char buffer/array that is big enough, no need for messing around with re-initialising memory like this; defining `char mybuffer[17]` in the header of an object in c++ automatically initialises it first time, and `sprintf` will add a terminating `0` so no need to clear it every loop.

GWToolboxdll/Widgets/SkillbarWidget.cpp
```diff
@@ -26,19 +26,18 @@ namespace
...
             return ms_to_string_sec(arr, cd);
         if (cd > 0s)
             return ms_to_string_secf(arr, cd);
-        if (cd == 0s)
-            arr[0] = '\0';
-        else // cd < 0, happens occasionally due to ping
-            strncpy_s(arr.data(), sizeof(arr), "0.0", 4); // avoid negative numbers in overlay
     }
 
     auto get_longest_effect_duration(GW::Constants::SkillID const skillId) -> auto
```
3vcloud, line 43: auto is great and all, but no need for defining the return types of functions. I have to read the function definition to figure out what its returning exactly instead of just looking at the declaration

GWToolboxdll/Widgets/SkillbarWidget.cpp
```diff
@@ -67,16 +79,33 @@ void SkillbarWidget::Draw(IDirect3DDevice9 *)
 
         for (auto it = 0u; it < 8; it++) {
             skill_cooldown_to_string(m_skills[it].cooldown, std::chrono::milliseconds{skillbar->skills[it].GetRecharge()});
-
-            auto const effect_duration = get_longest_effect_duration(static_cast<GW::Constants::SkillID>(skillbar->skills[it].skill_id));
+            auto const skill_id = static_cast<GW::Constants::SkillID>(skillbar->skills[it].skill_id);
+            auto const effect_duration = get_longest_effect_duration(skill_id);
```
3vcloud, line 83: Another example of auto - what is this return type without checking the function definition?

GWToolboxdll/Widgets/SkillbarWidget.cpp
```diff
@@ -110,6 +182,8 @@ void SkillbarWidget::LoadSettings(CSimpleIni *ini)
     m_skill_height = static_cast<int>(ini->GetLongValue(Name(), "height", 50));
     m_skill_width = static_cast<int>(ini->GetLongValue(Name(), "width", 50));
```
3vcloud, line 183: Don't initialize setting defaults twice; do it once via initialisation in the header file e.g. `m_skill_height = static_cast<int>(ini->GetLongValue(Name(), "height", m_skill_height ));`

GWToolboxdll/Widgets/SkillbarWidget.h
```diff
@@ -46,7 +47,9 @@ class SkillbarWidget final : public ToolboxWidget
 
     int m_skill_width = 50;
```
3vcloud, line 48: You could look into damage monitor or bonds monitor to have a good default depending on the player's choice of UI scaling.

GWToolboxdll/Widgets/SkillbarWidget.cpp
```diff
@@ -26,19 +26,18 @@ namespace
 
     auto ms_to_string_secf(std::array<char, 16> &arr, std::chrono::milliseconds const cd, const char *fmt = "%.1f") -> void
     {
-        snprintf(arr.data(), sizeof(ARRAYDESC), fmt, std::chrono::duration_cast<std::chrono::duration<float>>(cd).count());
+        snprintf(arr.data(), sizeof(arr), fmt, std::chrono::duration_cast<std::chrono::duration<float>>(cd).count());
```
3vcloud, line 29: Don't need to use `std::chrono` for this project; keep it simple and use the `TIMER_INIT()` and `TIMER_DIFF()` macros that have been defined for this stuff

GWToolboxdll/Widgets/SkillbarWidget.cpp
DubbleClick: The previous complaint from @HasKha was that std::string moves were too costly (everything on the heap due to SSO) so I changed it to directly modify the strings by reference
I don't see a single reason for why I should use a C-Style buffer array here just to truncate to pointers and then run into safety issues across functions. std::array has zero overhead.

The ZeroMemory was just left in from debugging purposes.

GWToolboxdll/Widgets/SkillbarWidget.cpp
DubbleClick: typing doesn't make me happy so `-> std::vector<std::chrono::milliseconds>` didn't seem too enticing to me, but I've changed it all across

GWToolboxdll/Widgets/SkillbarWidget.cpp
DubbleClick: I don't really see a benefit of using clock_t over std::chrono which was designed for the task, but sure.

GWToolboxdll/Widgets/SkillbarWidget.cpp
DubbleClick: I don't see a reason why anyone would need to know the exact type when we know it's a duration, but I've changed it to clock_t for whatever benefit

GWToolboxdll/Widgets/SkillbarWidget.h
DubbleClick: there's no info on the size of skills or the skill bar in total and I have no way of checking without resettings my interface back to the default, so I'll leave this be. This is rather little benefit just to save a very few players a tiny, tiny bit of customization. They need to move it to the correct position and configure if it's vertical or horizontal, anyway.

DubbleClick:
Showcase (almost everything enabled):
image

BearsMan:
awesome skill bar timer @DubbleClick love it

DubbleClick:
rebased to include the change to the infowindow

PR #502, opened 2020-08-16
fix and add some functionality to hero flagging hotkey

DubbleClick:


3vcloud:
The minimap setting shouldn't be in here, angle doesn't need validation. Other validation bits look good

DubbleClick:
Makes it a lot more understandable to the user I would think, to have a slider ranging from 0-360. There's no further validation done, than that, you can enter different values and it works.

Why shouldn't the minimap setting be in there, though? That was the functionality I actually expected.

3vcloud:
With 0 being right in front of the player I actually tested this using negative angles for positioning to the left, if anything it should be -180 to 180 or something but even then any negative or positive value is fine for this so no need. There is already a chat command to flag a hero to a location, and while this feature is different because its relative to an agent, it would be better suited for that chat command instead.

Features like this should be chat command orientated anyway in this same way that flagging a hero is, I just merged it in because it was cool and didn't consider that a hotkey to flag on minimap is also a nice to have

DubbleClick:
There's no chat command to trigger flagging on the minimap though, is there? I only had a brief look but it parsed x and y from the arguments iirc.

GWToolboxdll/Windows/Hotkeys.cpp
```diff
@@ -1192,32 +1199,36 @@ void HotkeyFlagHero::Description(char *buf, size_t bufsz) const
 }
 void HotkeyFlagHero::Draw()
 {
-    hotkeys_changed |= ImGui::InputFloat("Degree", &degree, 0.0f, 0.0f, 3);
-    hotkeys_changed |= ImGui::InputFloat("Distance", &distance, 0.0f, 0.0f, 3);
+    hotkeys_changed |= ImGui::DragFloat("Degree", &degree, 0.0f, -360.0f, 360.f);
```
3vcloud, line 1202: -180.0f to 180.0f

DubbleClick: But what if I expect 270 degrees to turn me 90 degrees counter clockwise?

PR #515, opened 2020-08-29
Color safe enemies

JaborGW:
When playing the t2 role in underworld, judging whether a behemoth behind the monument is safe to be left alive is often difficult and can easily cause a run to fail. From testing, it seems the safe area is the one between the two white lines:
safearea

This however, is often too small to judge in an actual run since the minimap is too cluttered. This patch gives users the option to instead give the agents of the "safe enemies" a different color.
This also applies the same to Dead Threshers and Collectors in the top group in pits. If this attracts interest, there are a few other cases this could be applied to, like the groups during the griffons quest in FoW.

### DIFF
--- a/GWToolboxdll/Widgets/Minimap/AgentRenderer.cpp
+++ b/GWToolboxdll/Widgets/Minimap/AgentRenderer.cpp
@@ -550,7 +550,7 @@ namespace
     bool isSafeThresher(float x, float y)
     {
         bool isSafe = (x - 12813.f) * (x - 12813.f) + (y - 4696.f) * (y - 4696.f) > 1012.5f * 1012.5f;
-        bool isRelevant = std::abs(13588.9f - x) + std::abs(4248.4f - y) < 500;
+        bool isRelevant = x > 13400 && x < 14500 && y > 3700 && y < 4700;
         return isRelevant && isSafe;
     }
 }
--- a/GWToolboxdll/Widgets/Minimap/AgentRenderer.cpp
+++ b/GWToolboxdll/Widgets/Minimap/AgentRenderer.cpp
@@ -544,8 +544,8 @@ namespace
 {
     bool isSafeBehemoth(float x, float y)
     {
-        auto isLeftOfLine = [x, y](std::tuple<float, float> a, std::tuple<float, float> b) { return ((std::get<0>(b) - std::get<0>(a)) * (y - std::get<1>(a)) - (std::get<1>(b) - std::get<1>(a)) * (x - std::get<0>(a))) > 0; };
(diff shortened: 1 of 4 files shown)


DubbleClick:
Definitely a great idea for stationary skele groups in fow, too.

JaborGW:
> Isn't that one too many curly brace?

The outer ones is for surrounding the constructor arguments, the middle ones for the initializer list and then each point. It does compile with one less only if each Vec2f is constructed explicitly (i.e. GW:Vec2f(x,y)); not exactly sure why.

DubbleClick:
Oh you're not assigning with an =, didn't pay attention.

3vcloud:
Rather than making custom code for certain agents in certain places for certain gameplay tactics, it would be better to just have a new option to draw polygons on the minimap in a similar way to how we can draw circles and lines. 

You wouldn't care about the colour of the agent if they were inside a big green polygon, and you wouldn't have to mess around with the agent renderer this way and suddenly you've made a feature that you don't have to explicitly code for every situation.

3vcloud:
Also please create PR's for dev branch in future

JaborGW:
> Rather than making custom code for certain agents in certain places for certain gameplay tactics, it would be better to just have a new option to draw polygons on the minimap in a similar way to how we can draw circles and lines.

I agree its not a very pretty solution, but this actually has decently big gameplay relevancy: Since groups of static enemies often spawn near exactly on top of these lines, even a few GWInches can make a difference here. This means any solution that relies on a player being able to judge this by position on the minimap himself will have limited usefulness.

> Also please create PR's for dev branch in future

Sorry for that, will do.

DubbleClick:
Unfortunately it would be very hard to see @3vcloud . Visual guidelines are great, like the aggro range too, but when enemies are very close to the lines or on top of them it's hard to judge if they're okay.

3vcloud:
Again this is way too niche to be used enough to be worth the code debt in any other area than the situation that you've described, when the line renderer does a good job of indicating this kind of thing; I use it in fow for griffs which works really well.

I can totally see this being podged for other areas and statements like "hey what if it patrols through this area, then thats a safe area and we should colour them green too at that point"

This is so niche its almost worth you making a new GWCA application just to handle it.

DubbleClick:
How about we instead allow colouring agents based on whether they're inside a custom polygon? That way we can get an accurate judgement but aren't required to hard code anything.

JaborGW:
That seems like a more elegant solution

3vcloud:
If this colour modifier is something you can toggle on or off for a custom polygon or circle on the map then that would work yep

DubbleClick:
Of course, I'll get to it later today or tomorrow.

PR #518, opened 2020-09-02
custom polygon (minimap) drawing & colouring of enemies inside those polygons

DubbleClick:
can get really freaky like this: 
image

or keep it tame with 
image

not filled
image

custom colours for enemies inside
image

settings overview
image


DubbleClick:
this implements #518 and #515

Hour-of-the-Owl:
I see how this is useful for mountains pop (that shape should be included by default), but what else is it good for? I'm just trying to think if there are any ways to improve this without interfering with existing use cases. For example, you could maybe have it remember the agent IDs of the mobs that spawned inside the shape, and not re-colour any that leave or enter.

DubbleClick:
Pits in uw, fow skeles in the path of griffs, derv or whatever the fuck lord in doa veil and many more

PR #544, opened 2020-10-13
don't update when running newer version

DubbleClick:


3vcloud:
1. Running in debug mode specifically turns off auto updating
2. If a release needs to be rolled back (i.e. delete a release in Github), doing this would prevent the updater from reverting the version.

PR #556, opened 2020-11-30
changed timer to not reset when zoning between explorables

HasKha:
changed timer to not reset when zoning between explorables. 

TO-DO:
- [x] figure out how to reset when zoning from normal explorable to dungeon (common for eotn dungeons). 
- [x] Call `SetRunCompleted()` from the objective timer once all objectives are completed to stop the timer.
- [x] Check that the timing is accurate and consistent with loading screen appearing. Maybe hook a different even to start the timer.


PostCrafter:
This looks really good already.
>  Check that the timing is accurate and consistent with loading screen appearing. Maybe hook a different even to start the timer.
I guess this is related to slow loading? Ideally I would like it to be synchronized between players, no matter their load time. This is especially useful for synchronizing actions between players ("use Ebon Escape at X:XX"). Maybe just offsetting the time by the instance time when they load in is an option?

>  figure out how to reset when zoning from normal explorable to dungeon (common for eotn dungeons).
I suggest adding a command that just resets the timer the next time you zone. This will be a lot more flexible than hard coding MapIDs

HasKha:
> I guess this is related to slow loading? Ideally I would like it to be synchronized between players, no matter their load time. This is especially useful for synchronizing actions between players ("use Ebon Escape at X:XX"). Maybe just offsetting the time by the instance time when they load in is an option?

I'll clarify. I want the timer to be the same across all the players in an instance, but also it should be possible for a non-tb player to have the same timer by adding an overlay afterwards. So I think we want to time since the first frame of loading screen. Different load times have no impact, as they all start loading at the same time. What I am not sure about, is if the current implementation indeed starts timing at the loading screen. I used the existing hook on `GW::Packet::StoC::GameSrvTransfer` and we might want to use a different one. 

> I suggest adding a command that just resets the timer the next time you zone. This will be a lot more flexible than hard coding MapIDs

Livia pointed me at `AreaInfo`, that should solve it. 
https://github.com/GregLando113/GWCA/blob/master/Include/GWCA/Managers/MapMgr.h#L65
https://github.com/GregLando113/GWCA/blob/master/Include/GWCA/GameEntities/Map.h#L41

PR #568, opened 2021-01-04
let users set default agent shape (minimap)

DubbleClick:
shapes will switch on map load, I'd assume that's fine since it's touched once and then likely never again

### DIFF
--- a/GWToolboxdll/Widgets/Minimap/AgentRenderer.cpp
+++ b/GWToolboxdll/Widgets/Minimap/AgentRenderer.cpp
@@ -60,6 +60,7 @@ void AgentRenderer::LoadSettings(CSimpleIni* ini, const char* section) {
     size_item = (float)ini->GetDoubleValue(section, VAR_NAME(size_item), 25.0);
     size_boss = (float)ini->GetDoubleValue(section, VAR_NAME(size_boss), 125.0);
     size_minion = (float)ini->GetDoubleValue(section, VAR_NAME(size_minion), 50.0);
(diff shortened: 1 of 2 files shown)


GWToolboxdll/Widgets/Minimap/AgentRenderer.cpp
```diff
@@ -60,6 +60,7 @@ void AgentRenderer::LoadSettings(CSimpleIni* ini, const char* section) {
     size_item = (float)ini->GetDoubleValue(section, VAR_NAME(size_item), 25.0);
     size_boss = (float)ini->GetDoubleValue(section, VAR_NAME(size_boss), 125.0);
     size_minion = (float)ini->GetDoubleValue(section, VAR_NAME(size_minion), 50.0);
+    default_shape = (Shape_e)ini->GetLongValue(section, VAR_NAME(default_shape), Circle)
```
3vcloud, line 63: From teardrop to circle by default?

DubbleClick: Since it's becoming increasingly common to lay the toolbox minimap over the game minimap, sure. I can revert this to tear though, I don't mind.

3vcloud: You've got tear set already  in the header file for this, but I would suggest not changing from the default that is already in toolbox so stick with that

DubbleClick: done

PR #585, opened 2021-04-11
Added Pre-Searing Trade chat to Trade Window 

Zedilurk:
Trade chat window now supports Pre-Searing and Kamadan chat via a button toggle at the top of the window.

### DIFF
--- a/GWToolboxdll/Windows/TradeWindow.cpp
+++ b/GWToolboxdll/Windows/TradeWindow.cpp
@@ -296,7 +296,7 @@ void TradeWindow::fetch() {
 
         // Check alerts
         // do not display trade chat while in kamadan AE district 1 or Pre-Searing Ascalon AE district 1
-        bool print_message = (print_game_chat && !GetInKamadanAE1() || print_game_chat_asc && !GetInAscalonAE1()) && IsTradeAlert(msg.message);
+        bool print_message = ((is_kamadan_chat && print_game_chat && !GetInKamadanAE1()) || (!is_kamadan_chat && print_game_chat_asc && !GetInAscalonAE1())) && IsTradeAlert(msg.message);
 
         if (print_message) {
             wchar_t buffer[512];
--- a/GWToolboxdll/Windows/TradeWindow.cpp
+++ b/GWToolboxdll/Windows/TradeWindow.cpp
@@ -580,9 +580,11 @@ void TradeWindow::RegisterSettingsContent()
 void TradeWindow::DrawAlertsWindowContent(bool) {
     ImGui::Text("Alerts");
     ImGui::Checkbox("Send Kamadan AE1 trade chat to your trade chat", &print_game_chat_kmd);
+    ImGui::ShowHelp("Only when trade chat channel is visible in-game");
     ImGui::Checkbox("Send Pre-Searing Ascalon AE1 trade chat to your trade chat", &print_game_chat_asc);
     ImGui::ShowHelp("Only when trade chat channel is visible in-game");
     ImGui::Checkbox("Only show messages containing:", &filter_alerts);
+    ImGui::ShowHelp("Only shows messages from the currently active trade channel (Kamadan OR Ascalon)");
     ImGui::TextDisabled("(Each line is a separate keyword. Not case sensitive.)");
     if (ImGui::InputTextMultiline("##alertfilter", alert_buf, ALERT_BUF_SIZE, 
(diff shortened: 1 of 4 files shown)


3vcloud:
This is good, but needs a hover tooltip on the button to explain what it does, and refactor in the alerts section (this would work with Ascalon messages, not not when you're in Kamadan AE1 for example)

Zedilurk:
- Added Tooltip to button as mentioned.
- Added support for both Ascalon and Kamadan alerts.
- Added notes to indicate that only alerts from the active channel will be sent to your trade chat since there is only one active websocket at the moment (not one for each channel).

GWToolboxdll/Windows/TradeWindow.cpp
```diff
@@ -286,8 +295,8 @@ void TradeWindow::fetch() {
             messages.add(msg);
 
         // Check alerts
-        // do not display trade chat while in kamadan AE district 1
-        bool print_message = print_game_chat && !GetInKamadanAE1() && IsTradeAlert(msg.message);
+        // do not display trade chat while in kamadan AE district 1 or Pre-Searing Ascalon AE district 1
+        bool print_message = (print_game_chat_kmd && !GetInKamadanAE1() || print_game_chat_asc && !GetInAscalonAE1()) && IsTradeAlert(msg.message);
```
3vcloud, line 299: Check your operators here

GWToolboxdll/Windows/TradeWindow.cpp
```diff
@@ -591,7 +609,8 @@ void TradeWindow::DrawSettingInternal() {
 
 void TradeWindow::LoadSettings(CSimpleIni* ini) {
     ToolboxWindow::LoadSettings(ini);
-    print_game_chat = ini->GetBoolValue(Name(), VAR_NAME(print_game_chat), print_game_chat);
+    print_game_chat_kmd = ini->GetBoolValue(Name(), VAR_NAME(print_game_chat), print_game_chat_kmd);
```
3vcloud, line 614: Settings key is the same, need to tweak this. Keep "print_game_chat" as the kamadan one so the update doesn't mess with users' settings

GWToolboxdll/Windows/TradeWindow.cpp
```diff
@@ -608,7 +627,8 @@ void TradeWindow::LoadSettings(CSimpleIni* ini) {
 void TradeWindow::SaveSettings(CSimpleIni* ini) {
     ToolboxWindow::SaveSettings(ini);
 
-    ini->SetBoolValue(Name(), VAR_NAME(print_game_chat), print_game_chat);
+    ini->SetBoolValue(Name(), VAR_NAME(print_game_chat), print_game_chat_kmd);
```
3vcloud, line 632: Settings key is the same, need to tweak this. Keep "print_game_chat" as the kamadan one so the update doesn't mess with users' settings

GWToolboxdll/Windows/TradeWindow.cpp
Zedilurk: I believe the operators are correct here in that it checks both cities and the settings specific to that city and will print if either is correct.

However, I do see an edge case where if the user is connected to Kamadan chat, has 'print_game_chat_asc' turned on, and is in Kamadan AE1 and has alerts set, it would be reading the Kamadan message queue and printing while in Kamadan still because it determined they aren't in Ascalon, which is what they actually have set for their settings and I don't believe messages have any indicator on where they originate from.

That fix will definitely require a couple lines though, so I'll give that a look.

Zedilurk:
Accidentally committed a change I was looking at via the friends list, so I reverted that.

-Fixed issue with overlapping Settings keys.
-Reverted variable to prevent messing with User's Settings.

Looking into an edge case regarding the 'print_message' line mentioned above.

GWToolboxdll/Windows/TradeWindow.cpp
3vcloud: Nope, just use more brackets

Zedilurk: Ahh I see what you mean, fixed. Also fixed the edge case I had mentioned by checking which city chat we are connected to.

Zedilurk:
Fixed OR operator and an edge case where having 'Print Ascalon' on while in/and connected to Kamadan (ie, fullfilling the !AscalonAE1 requirement) would print message alerts from our connected websocket (Kamadan).

Issue #541, opened 2020-10-11
Distance Widget - Customizable Display Options


T0rv4ld:
Distance widget offers a couple of options related to when and where to display it, however it lacks options to turn ON/OFF the display of its components. Currently when displayed, it contains 3 parts:
- The word Distance on the first line (always white)
- The percentage based on how far/close we are from the target on the second line (color customizable)
- The actual in-game distance on the 3rd line (always white)

It is currently not possible to get rid of any of the lines above ; allowing via tickboxes **each of those 3 components** to be turn ON/OFF would allow to display, for example, only the percentage. When placing it near the minimap, this would confirm, using peripheral vision and proper color, that we are indeed in the right range of the target.

In my case, to take an example, I think the "distance" title is not required (as we place the widget so we know what it is for) and the in game number is only usable for specific scenarios, as it is too small, changes too fast and is not colored and therefore not of much use in a speed run to highlight the position. All I would need to tick on is the percentage display, as it is the only one colored and an easy number to remember (20-25% for cast range). Would make the overall on screen display and UI cleaner and less crowded.

PR #603, opened 2021-05-26
Distance Widget - Added Display Options

franneck94:
Added two display options for the distance widget that were also suggested (partially) by the [Issue](https://github.com/HasKha/GWToolboxpp/issues/541#issue-718904959)

- Enable/Disable the percentage distance value
- Enable/Disable the absolute distance value


3vcloud:
Merged into dev branch

PR #606, opened 2021-05-28
Update Hotkeys.cpp

assblaster9000:
Added gonpc if in explorable. Use case: Putting 8578561,8578567,8581889,8581895, all on 1 key allows you to accept the quest and take the reward for both, Breaching the Stygian Veil, and Brood Wars. (If the quest's are completed) . Without gonpc you'd have to hit the same hotkey 4 times to achieve the same results.


3vcloud:
Merged into dev, please PR on that branch for next time

PR #609, opened 2021-06-01
Write Party Damage by Clicking

franneck94:
By clicking on the party damage window the damage is written to the party chat.
This would be an alternative to the chat command.

Note: I am not sure if i should add an option for this to enable/disable.


HiImFlo:
Of course this should disabled by default. What you probably would want to do is only ping on Ctrl+Click, like the health monitor does.

3vcloud:
1. Should be ctrl + click as @HiImFlo mentioned
2. Should only write the current row to chat instead of triggering a mess of messages that you can't stop if you've accidentally ctrl + clicked in that area

3. Also please keep functional commits separate from coding style commits so I can more easily see whats what :)

franneck94:
1.) Of course, I will implement this today.
2.) So you mean writing only the damage of a single player?

3vcloud:
Yep, the player can ctrl click each row themselves if they really want to

franneck94:
1.) and 2.) are now implemented

PR #622, opened 2021-07-31
added targeting by aget id

HasKha:
added an optional 2nd argument to /target id index, to target based on index-th agent id

this is useful when different party member need to target different enemies with the same model ids

### DIFF
--- a/GWToolboxdll/Modules/ChatCommands.cpp
+++ b/GWToolboxdll/Modules/ChatCommands.cpp
@@ -283,7 +283,7 @@ void ChatCommands::DrawHelp() {
     ImGui::Bullet(); ImGui::Text("'/target closest' to target the closest agent to you.\n"
         "'/target ee' to target best ebon escape agent.\n"
         "'/target hos' to target best vipers/hos agent.\n"
-        "'/target [name|model_id]' target nearest NPC by name or model_id.\n"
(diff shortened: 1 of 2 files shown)


GWToolboxdll/Modules/ChatCommands.cpp
```diff
@@ -1320,11 +1318,20 @@ bool ChatCommands::GetTargetTransmoInfo(PendingTransmo &transmo)
...
+
     // Searching by name; offload this to decode agent names first.
-    if (!GuiUtils::ParseUInt(model_id_or_name, &model_id)
-        && !IsNearestStr(model_id_or_name)) {
-        Instance().npc_to_find.Init(model_id_or_name,type);
-        return;
+    if (GuiUtils::ParseUInt(model_id_or_name, &model_id)) {
+        // check if there's an index component
+        if (const wchar_t* rest = GetRemainingArgsWstr(model_id_or_name, 1)) {
+            GuiUtils::ParseUInt(rest, &index);
```
3vcloud, line 1327: Does this only work when targeting by model_id, not by name?

HasKha: yea, the usage is so niche already, I didn't bother adding that. And it's probably not worth the extra complexity.

PR #646, opened 2022-01-08
Travel to a specific disctrict

chypriote:
Allows the /tp command to handle specifying a district number
i.e: `/tp kama ae2` will travel to Kamadan American District 2

Note that I am not that experienced with C++ and my way to parse the district number is probably very wrong. The way I did, it  wont work for districts >9 (this should only happen during events like wintersday/boardwalk, so shouldn't be much of a problem).    
If you have a better way of parsing the district number I'd be glad to update this PR.

Side question, I noticed the code to travel to a specific Guild Hall was removed, is there a reason ?

### DIFF
--- a/GWToolboxdll/Windows/TravelWindow.cpp
+++ b/GWToolboxdll/Windows/TravelWindow.cpp
@@ -1432,11 +1432,10 @@ bool TravelWindow::ParseDistrict(const std::wstring &s, GW::Constants::District 
 {
     std::string compare = GuiUtils::ToLower(GuiUtils::RemovePunctuation(GuiUtils::WStringToString(s)));
     std::string first_word = compare.substr(0, compare.find(' '));
+    char alias[3];
 
-    // Parse the district number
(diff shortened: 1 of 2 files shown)


3vcloud:
This looks like a good fit for using a regex to find the district alias and number; do some googling on it and see https://github.com/HasKha/GWToolboxpp/blob/dde1f73fa956d1bd291c8ba0c6371926629a2c42/GWToolboxdll/Modules/GameSettings.cpp#L440 for an example of how its used.

Traveling to other GH's can be exploited for pvp so was removed from tb

reduf:
Might be a bit cleaner to use `sscanf` instead of a regex.
```c
unsigned number;
char alias[3];
sscanf(s, "%2s%u", &alias, &number);
```

chypriote:
@reduf thanks for this, I've been trying on and off to get this to work with regexes without success.  
sscanf seems to be working perfectly.

3vcloud:
What about int?

Issue #535, opened 2020-09-30
Hero skill usage statistics


quetzalsly:
It would be really cool to have a panel that describes each skill that each hero is using (Kinda like arcdps does in GW2).
Things like frequency of usage, % of total damage, energy at time of cast and so on.
This would really help people theorycrafting builds and answer questions like "Will my mesmers do more damage if I take out spiritual pain, will they cast energy surge more in that case?".

PR #654, opened 2022-02-22
Feat: Party Skill Statistics

franneck94:
Hey,
I've added a new Window for the statistics of the skill usage of the team members, as suggested in #535.

The new feature consist of the following:

- Get the names of the players and their skills:
  - Load the skillbar for the player and his heroes
  - For other players or henchmen the skillbar cannot be loaded, therefore the skills are initialized with "Unknown Skill" 
   - When entering an explorable and the other player/henchman will activate their skills, these will then replace the beforehand "Unknown Skills"
- After leaving an explorable and going back to the outpost, the data from the previous run will be displayed until any party member leaves or a new party member joins

Example:

Screenshot_8

For every player/hero/henchman in the party:
- The absolute number of activations for each skill, if the setting for displaying the absolute number is active
- The percentage of activations for each skill, if the setting for displaying the percentage value is active

These stats can be posted in the chat by the following commands:
- /skillstats: Will display the stats for the player
- /skillstats 1: Will display the stats for the team member at position 1
- /skillstats 1 1: Will display the stats for the first skill of the first team member
- /skillstats reset: Will reset the statistics

Note: Skills that were not activated yet (abs. count of 0) will not be printed in the chat.

Screenshot_4

One alternative to posting the info in the chat is by pressing Ctrl and Left Clicking on the player's stats in the party statistics window.

### DIFF
--- a/GWToolboxdll/Modules/ChatCommands.cpp
+++ b/GWToolboxdll/Modules/ChatCommands.cpp
@@ -1745,7 +1745,7 @@ void ChatCommands::CmdSkillStatistics(const wchar_t* message, int argc, LPWSTR* 
     if (argc == 2) {
         /* command: /skillstats reset */
         if (arg1 == L"reset") {
-            PartyStatisticsWindow::Instance().ResetPlayerStatistics();
+            PartyStatisticsWindow::Instance().UnsetPlayerStatistics();
         }
         /* command: /skllstats playerNum */
         else {
--- a/GWToolboxdll/Modules/ChatCommands.cpp
+++ b/GWToolboxdll/Modules/ChatCommands.cpp
@@ -278,7 +278,7 @@ void ChatCommands::DrawHelp() {
     ImGui::Bullet(); ImGui::Text("'/flag [all|clear|<number>]' to flag a hero in the minimap (same as using the buttons by the minimap).");
     ImGui::Bullet(); ImGui::Text("'/flag [all|<number>] [x] [y]' to flag a hero to coordinates [x],[y].");
     ImGui::Bullet(); ImGui::Text("'/flag <number> clear' to clear flag for a hero.");
-
+    
     ImGui::Bullet(); ImGui::Text("'/load [build template|build name] [Hero index]' loads a build. The build name must be between quotes if it contains spaces. First Hero index is 1, last is 7. Leave out for player");
     ImGui::Bullet(); ImGui::Text("'/pcons [on|off]' toggles, enables or disables pcons.");
     const char* toggle_hint = "<name> options: helm, costume, costume_head, cape, <window_or_widget_name>";
@@ -350,30 +350,30 @@ void ChatCommands::Initialize() {
     const DWORD def_scale = 0x64000000;
     // Available Transmo NPCs
     // @Enhancement: Ability to target an NPC in-game and add it to this list via a GUI
-    npc_transmos = {
-        {"charr", {163, def_scale, 0x0004c409, 0, 98820}},
-        {"reindeer", {5, def_scale, 277573, 277576, 32780}},
-        {"gwenpre", {244, def_scale, 116377, 116759, 98820}},
-        {"gwenchan", {245, def_scale, 116377, 283392, 98820}},
+    npc_transmos = { 
+        {"charr", {163, def_scale, 0x0004c409, 0, 98820}}, 
+        {"reindeer", {5, def_scale, 277573, 277576, 32780}}, 
+        {"gwenpre", {244, def_scale, 116377, 116759, 98820}}, 
+        {"gwenchan", {245, def_scale, 116377, 283392, 98820}}, 
         {"eye", {0x1f4, def_scale, 0x9d07, 0, 0}},
-        {"zhu", {298, def_scale, 170283, 170481, 98820}},
-        {"kuunavang", {309, def_scale, 157438, 157527, 98820}},
-        {"beetle", {329, def_scale, 207331, 279211, 98820}},
-        {"polar", {313, def_scale, 277551, 277556, 98820}},
-        {"celepig", {331, def_scale, 279205, 0, 0}},
-        {"mallyx", {315, def_scale, 243812, 0, 98820}},
-        {"bonedragon", {231, def_scale, 16768, 0, 0}},
-        {"destroyer", {312, def_scale, 285891, 285900, 98820}},
-        {"destroyer2", {146, def_scale, 285886, 285890, 32780}},
-        {"koss", {250, def_scale, 243282, 245053, 98820}},
-        {"smite", {346, def_scale, 129664, 0, 98820}},
-        {"dorian", {8299, def_scale, 86510, 0, 98820}},
-        {"kanaxai", {317, def_scale, 184176, 185319, 98820}},
-        {"skeletonic", {359, def_scale, 52356, 0, 98820}},
+        {"zhu", {298, def_scale, 170283, 170481, 98820}}, 
+        {"kuunavang", {309, def_scale, 157438, 157527, 98820}}, 
+        {"beetle", {329, def_scale, 207331, 279211, 98820}},     
+        {"polar", {313, def_scale, 277551, 277556, 98820}},    
+        {"celepig", {331, def_scale, 279205, 0, 0}},  
+        {"mallyx", {315, def_scale, 243812, 0, 98820}},     
+        {"bonedragon", {231, def_scale, 16768, 0, 0}},     
+        {"destroyer", {312, def_scale, 285891, 285900, 98820}},   
+        {"destroyer2", {146, def_scale, 285886, 285890, 32780}},    
+        {"koss", {250, def_scale, 243282, 245053, 98820}},     
+        {"smite", {346, def_scale, 129664, 0, 98820}}, 
+        {"dorian", {8299, def_scale, 86510, 0, 98820}},         
+        {"kanaxai", {317, def_scale, 184176, 185319, 98820}},         
+        {"skeletonic", {359, def_scale, 52356, 0, 98820}},          
         {"moa", {504, def_scale, 16689, 0, 98820}}
     };
 
-    // you can create commands here in-line with a lambda, but only if they are only
+    // you can create commands here in-line with a lambda, but only if they are only 
     // a couple of lines and not used multiple times
     GW::Chat::CreateCommand(L"ff", [](const wchar_t*, int, LPWSTR*) {
         GW::Chat::SendChat('/', "resign");
@@ -443,7 +443,7 @@ bool ChatCommands::WndProc(UINT Message, WPARAM wParam, LPARAM lParam) {
                 case VK_W:
                 case VK_X:
                 case VK_Z:
-
+                
                 case VK_ESCAPE:
                 case VK_UP:
                 case VK_DOWN:
@@ -460,8 +460,8 @@ void ChatCommands::Update(float delta) {
 
     if (delta == 0.f) return;
 
-    if (GW::CameraMgr::GetCameraUnlock()
-        && !GW::Chat::GetIsTyping()
+    if (GW::CameraMgr::GetCameraUnlock() 
+        && !GW::Chat::GetIsTyping() 
         && !ImGui::GetIO().WantTextInput) {
 
         float forward = 0;
@@ -675,8 +675,8 @@ void ChatCommands::CmdEnterMission(const wchar_t*, int argc, LPWSTR* argv) {
     const char* const error_fow_uw_syntax = "Use '/enter fow' or '/enter uw' to trigger entry";
     const char* const error_no_scrolls = "Unable to enter elite area; no scroll found";
     const char* const error_not_leading = "Unable to enter mission; you're not party leader";
-
-    if (GW::Map::GetInstanceType() != GW::Constants::InstanceType::Outpost)
+    
+    if (GW::Map::GetInstanceType() != GW::Constants::InstanceType::Outpost) 
         return Log::Error(error_use_from_outpost);
 
     switch (GW::Map::GetMapID()) {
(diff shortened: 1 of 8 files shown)


GWToolboxdll/Windows/PartyStatistics.cpp
```diff
@@ -346,11 +335,19 @@ void PartyStatisticsWindow::Initialize() {
 
     ToolboxWindow::Initialize();
 
+    /* Reset party data on player add */
+    GW::StoC::RegisterPacketCallback<GW::Packet::StoC::PartyPlayerAdd>(&PartyPlayerAdd_Callback,
```
3vcloud: Instead of having lambda function, declare `PartyClearCallback` as static with the relevent arguments.

GWToolboxdll/Windows/PartyStatistics.cpp
```diff
@@ -24,11 +24,11 @@
 void PartyStatisticsWindow::GetSkillName(const uint32_t skill_id, char* const skill_name) {
     const auto& skill_data = GW::SkillbarMgr::GetSkillConstantData(skill_id);
 
-    const auto length = size_t{32};
-    wchar_t buffer[length] = {L'\0'};
+    const auto buffer_length = size_t{32};
```
3vcloud: Don't use autos unless its actually useful, just declare it as `const size_t buffer_length = 32`

GWToolboxdll/Modules/ChatCommands.cpp
```diff
@@ -421,7 +421,7 @@ void ChatCommands::Initialize() {
     });
     GW::Chat::CreateCommand(L"hero", ChatCommands::CmdHeroBehaviour);
     GW::Chat::CreateCommand(L"morale", ChatCommands::CmdMorale);
-    GW::Chat::CreateCommand(L"morale", ChatCommands::CmdSkillStatistics);
+    GW::Chat::CreateCommand(L"skillstats", ChatCommands::CmdSkillStatistics);
```
3vcloud, line 424: Put this into your module, not in `ChatCommands` - modules can be disabled so this would be a problem

GWToolboxdll/Windows/PartyStatistics.cpp
```diff
@@ -0,0 +1,467 @@
...
+    if (nullptr == buffer) return;
+
+    const auto length = size_t{32};
+    GW::UI::AsyncDecodeStr(buffer, agent_name, length);
+}
+
+void PartyStatisticsWindow::ClearOnPartySizeChange() {
+    if (GW::Constants::InstanceType::Outpost != GW::Map::GetInstanceType()) return;
+
+    static auto last_party_size = static_cast<size_t>(-1);
```
3vcloud: no auto

GWToolboxdll/Windows/PartyStatistics.cpp
```diff
@@ -0,0 +1,467 @@
...
+    const auto party_member_found = agent_it != Instance().party_stats.end();
+
+    if (!party_member_found) return;
+
+    const auto agent_idx = std::distance(Instance().party_stats.begin(), agent_it);
+
+    auto& member_stats = Instance().party_stats[agent_idx];
+    auto& member_skill_counts = member_stats.skill_counts;
+
+    const auto skill_it = std::find_if(member_skill_counts.begin(), member_skill_counts.end(),
```
3vcloud: No idea what any of the var types are looking at this code without an IDE to hold my hand - try not to use auto in this way too much

GWToolboxdll/Windows/PartyStatistics.cpp
```diff
@@ -0,0 +1,467 @@
...
+            ImGui::Text("Hero data cannot be loded in outpost n/a");
+        }
+        return;
+    }
+
+    GetPlayerName(agent, agent_name);
+
+    snprintf(header_label, buffer_length, "%s", agent_name);
+
+    static auto is_open = true;
```
3vcloud: no auto

GWToolboxdll/Windows/PartyStatistics.cpp
```diff
@@ -25,19 +25,28 @@
...
 
-    wchar_t buffer[BUFFER_LENGTH] = {L'\0'};
+    wchar_t encoding_buffer[BUFFER_LENGTH] = {L'\0'};
+    wchar_t decoding_buffer[BUFFER_LENGTH] = {L'\0'};
 
     const auto encoding_length = size_t{16};
-    if (GW::UI::UInt32ToEncStr(skill_data.name, buffer, encoding_length)) {
-        GW::UI::AsyncDecodeStr(buffer, skill_name, BUFFER_LENGTH);
+    if (GW::UI::UInt32ToEncStr(skill_data.name, encoding_buffer, encoding_length)) {
+        GW::UI::AsyncDecodeStr(encoding_buffer, decoding_buffer, BUFFER_LENGTH);
```
3vcloud, line 36: Don't call this synchronously and expect it to work all the time. Use `GuiUtils::EncStr` which wraps this

GWToolboxdll/Modules/ChatCommands.cpp
```diff
@@ -349,30 +349,30 @@ void ChatCommands::Initialize() {
     const DWORD def_scale = 0x64000000;
     // Available Transmo NPCs
     // @Enhancement: Ability to target an NPC in-game and add it to this list via a GUI
-    npc_transmos = { 
-        {"charr", {163, def_scale, 0x0004c409, 0, 98820}}, 
-        {"reindeer", {5, def_scale, 277573, 277576, 32780}}, 
-        {"gwenpre", {244, def_scale, 116377, 116759, 98820}}, 
-        {"gwenchan", {245, def_scale, 116377, 283392, 98820}}, 
+    npc_transmos = {
```
3vcloud, line 353: Don't bother reformatting files that you haven't touched - its unnecessary chaff to check through without any functional change being made

3vcloud (commented):
TLDR; Don't use autos when it makes variables harder to identify the type, use static methods for some packet callbacks, don't clang format files that haven't changed functionally

GWToolboxdll/Windows/PartyStatistics.cpp
```diff
@@ -0,0 +1,467 @@
...
+        if (static_cast<uint32_t>(0U) == total_num_skills) total_num_skills = 1U;
+
+        for (const auto [skill_id, count] : party_member_stats.skill_counts) {
+            char skill_name[buffer_length] = {'\0'};
+            GetSkillName(skill_id, skill_name);
+
+            const auto percentage = (static_cast<float>(count) / static_cast<float>(total_num_skills)) * 100.0F;
+
+            if (show_perc_values && show_abs_values) {
+                ImGui::Text("%s: %u\t-\t%-3.2f%%", skill_name, count, percentage);
```
3vcloud: Align this in columns

GWToolboxdll/Windows/PartyStatistics.cpp
franneck94: Is there any special reason? Throughout the code base, I saw both.

GWToolboxdll/Modules/ChatCommands.cpp
franneck94: What exactly do you mean? My command function is kind of the same strucutre than for example the CmdDamage function.

GWToolboxdll/Windows/PartyStatistics.cpp
franneck94: Should this then also be updated for the following modules?
Or was it inteded there?

Screenshot_6

GWToolboxdll/Windows/PartyStatistics.cpp
franneck94: Done.

GWToolboxdll/Windows/PartyStatistics.cpp
franneck94: Done.

GWToolboxdll/Windows/PartyStatistics.cpp
franneck94: Done.

GWToolboxdll/Windows/PartyStatistics.cpp
franneck94: Done.

GWToolboxdll/Windows/PartyStatistics.cpp
3vcloud: If it doesn't break anything then a PR that has the dependency pulled up to date would be welcome, but I doubt it'll be that simple ;)

GWToolboxdll/Windows/PartyStatistics.cpp
3vcloud: There are some edge cases but yes, these lines will be refactored when I get around to it. The class I made makes it a bit more manageable, but you'll need to figure out how to keep a reference to these decoded strings across frames

GWToolboxdll/Modules/ChatCommands.cpp
3vcloud: If you were to disable the damage module in tb settings, restart toolbox and then run the damage command, it wouldnt work because the module isn't running. Its compiled in with toolbox so you won't crash or anything but keeping module functions as separate as you can means that if you add an optional module from another DLL at runtime its all self contained

GWToolboxdll/Windows/PartyStatistics.cpp
3vcloud: If you're not doing any logic in the lambda its pointless - depending on the compiler, you've got 2 jumps instead of 1 and the code would be more readable. Your generic packet handler callback is a good example of when its OK to use it because you're passing in different vars depending on the packet

GWToolboxdll/Windows/PartyStatistics.cpp
franneck94: For now i use the old Columns API.

Is this layout okay for you?

Screenshot_7

GWToolboxdll/Modules/ChatCommands.cpp
franneck94: Noted for the future. Should i revert it for now?

GWToolboxdll/Windows/PartyStatistics.cpp
3vcloud: Looks good yeah, I'd keep last 2 columns fixed with and the first can be whatever is left. Don't forget to scale with global font scale

GWToolboxdll/Modules/ChatCommands.cpp
3vcloud: Yep revert all of these files for now, makes it easier to track feature changes

GWToolboxdll/Windows/PartyStatistics.cpp
franneck94: So it isnt enough to just use the current window width?

```cpp
const float width = ImGui::GetWindowWidth();
ImGui::Columns(3, table_name, false);
ImGui::SetColumnWidth(0, width * 0.6F);
ImGui::SetColumnWidth(1, width * 0.2F);
ImGui::SetColumnWidth(2, width * 0.2F);
```

3vcloud: Guess so, I usually keep columns that won't contain string text to a fixed width I.e. when you expand the window, only the skill name fills the extra space but this works fine

GWToolboxdll/Modules/ChatCommands.cpp
franneck94: I still dont get your point. Are you refering to exactly this line in the review, or to the CmdSkillStatistics function itself?

3vcloud: This line needs to be in your module, not the chat commands module

franneck94: Okay i will do this.

But sorry for being stupid, but why is the GW::Chat::CreateCommand(L"dmg", ChatCommands::CmdDamage); command then in the ChatCommands.cpp and not in the PartyDamage.cpp?
Or is this just an outdated code style?

3vcloud: Yep, there's an ongoing task to allow modules to be side loaded into toolbox, which means the functionality of each module would be self contained. This includes registering hooks for chat commands

GWToolboxdll/Windows/PartyStatistics.cpp
```diff
@@ -25,28 +25,32 @@
...
+        skill_name_enc = (*it).second.encoder;
     }
 
-    std::wstring skill_name{decoding_buffer};
+    if (!skill_name_enc || skill_name_enc->IsDecoding()) return UNKNOWN_SKILL_NAME;
 
-    if (NONE_SKILL_NAME == skill_name) {
-        skill_name = std::wstring{UNKNOWN_SKILL_NAME};
-    }
+    const std::wstring result = skill_name_enc->wstring();
```
3vcloud, line 45: I think this would do a copy

GWToolboxdll/Windows/PartyStatistics.cpp
```diff
@@ -25,28 +25,32 @@
...
 
-    return skill_name;
+    if (result.empty()) return UNKNOWN_SKILL_NAME;
+
+    delete skill_name_enc;
```
3vcloud, line 49: You'll leave a hanging pointer if skill name hasn't decoded yet

franneck94: While a skill name is still decoding no new heap allocation will happen for this skill und the already created pointer will be used (logic is in the if/else at the beginning of the function).

GWToolboxdll/Windows/PartyStatistics.cpp
franneck94: Should i do a move after decoding is finished? A Reference wouldnt work since i will delete the pointer at the end of the function.

franneck94: Should i do a move after decoding is finished? A Reference wouldnt work since i will delete the pointer at the end of the function.

franneck94: Do you just mean i should only do a copy after decoding is finished?

GWToolboxdll/Windows/PartyStatistics.cpp
3vcloud: Did a quick amendment @ https://github.com/3vcloud/GWToolboxpp/commit/e3849bacf65cec2871be4df3c16dac21ae25f2e7 to make string decoding a little less confusing

PR #665, opened 2022-04-22
quick safety check before deleting old directory

DubbleClick:


3vcloud:
I've gone through the whole logic process and rewritten it to handle any edge cases that haven't been covered, so this PR can be closed.

May want to rebase onto dev

PR #678, opened 2022-05-12
added two faq that have been asked in the discord

Janitsu:
I have added a solution to the two issues; icons not loading and Font.ttf being missing. These are the solutions that, according to users that have had their issues rectified, are working. I even tried it myself on my installation. I am not sure if either of these are solved by using the solution provided in "The icons on the Pcons, Materials, and Toolbox window aren't showing up. How do I get them back?", If they are, then ignore this by all means 👍.


DubbleClick:
The toolbox folder has been moved to C:Users/<user>/(OneDrive/)Documents/GWToolboxpp/<Computername>

Janitsu:
I suppose this is a newer change and old installations will still follow the previous location? At least my settings folder (when opened through TB) lead me to that address. I will change that anyway for all mentions of the folder.

3vcloud:
Merge into dev

PR #695, opened 2022-06-07
Latency widget

backtwobaysix:


BearsMan:
this should be test in dev on master branch XD

PR #701, opened 2022-06-10
Agent borders on minimap

backtwobaysix:


GWToolboxdll/Widgets/Minimap/AgentRenderer.cpp
```diff
@@ -568,6 +573,10 @@ void AgentRenderer::Enqueue(const GW::Agent* agent, const CustomAgent* ca) {
         // Target highlight if this is the current or next target
         if (auto_target_id == agent->agent_id || GW::Agents::GetTargetId() == agent->agent_id)
             Enqueue(shape, agent, size + 50.0f, Colors::Sub(color_target, IM_COL32(0, 0, 0, 50)));
+        else {
+            if (agent_border)
+                Enqueue(shape, agent, size + agent_border_thickness, Colors::Sub(0xFF000000, IM_COL32(0, 0, 0, 50)));
```
DubbleClick, line 578: Why do you substract 50 alpha from static 255 alpha? Just use 0xCD000000

GWToolboxdll/Widgets/Minimap/AgentRenderer.cpp
```diff
@@ -61,6 +62,7 @@ void AgentRenderer::LoadSettings(CSimpleIni* ini, const char* section) {
     size_boss = (float)ini->GetDoubleValue(section, VAR_NAME(size_boss), 125.0);
     size_minion = (float)ini->GetDoubleValue(section, VAR_NAME(size_minion), 50.0);
     default_shape = (Shape_e)ini->GetLongValue(section, VAR_NAME(default_shape), default_shape);
+    agent_border_thickness = (int)ini->GetLongValue(section, VAR_NAME(agent_border_thickness), 20);
```
DubbleClick, line 65: uint please, thickness shouldn't be negative

GWToolboxdll/Widgets/Minimap/AgentRenderer.h
```diff
@@ -76,6 +78,8 @@ class AgentRenderer : public VBuffer {
         bool color_text_active = false;
         bool shape_active = true;
         bool size_active = false;
+        bool agent_border = true;
```
DubbleClick, line 81: why are these in CustomAgent too, do you plan to extend custom borders to custom agents? if not, this should be removed

DubbleClick (changes requested):

GWToolboxdll/Widgets/Minimap/AgentRenderer.cpp
backtwobaysix: So it can be adjusted easier, but I have changed it

GWToolboxdll/Widgets/Minimap/AgentRenderer.cpp
backtwobaysix: Fixed

GWToolboxdll/Widgets/Minimap/AgentRenderer.h
backtwobaysix: This is purely for the settings save/load, shall I try and move the setting into minimap

DubbleClick: they're in the correct place, but you've added them twice, once in AgentRenderer (where they belong) and once in CustomAgent (here) which are unused

backtwobaysix: Could be wrong but I'm pretty sure these are the values that agentrenderer:customagent::savesettings() uses. If you comment them out it stops working

BearsMan: @DubbleClick is right it should not be CustomAgent

DubbleClick: they shouldn't be in AgentRenderer::CustomAgent::SaveSettings, they should be saved and loaded in AgentRenderer::SaveSettings

BearsMan: ^^

backtwobaysix: Ah agentrenderer:customagent::savesettings() did nothing in the first place, fixed

DubbleClick:
👍

GWToolboxdll/Widgets/Minimap/AgentRenderer.cpp
```diff
@@ -61,6 +62,7 @@ void AgentRenderer::LoadSettings(CSimpleIni* ini, const char* section) {
     size_boss = (float)ini->GetDoubleValue(section, VAR_NAME(size_boss), 125.0);
     size_minion = (float)ini->GetDoubleValue(section, VAR_NAME(size_minion), 50.0);
     default_shape = (Shape_e)ini->GetLongValue(section, VAR_NAME(default_shape), default_shape);
+    agent_border_thickness = (uint32_t)ini->GetLongValue(section, VAR_NAME(agent_border_thickness), 20);
```
DubbleClick, line 65: this is what I mean, this should be `agent_border_thickness = (uint32_t)ini->GetLongValue(section, VAR_NAME(agent_border_thickness), agent_border_sickness);`

GWToolboxdll/Widgets/Minimap/AgentRenderer.h
```diff
@@ -26,6 +26,8 @@ class AgentRenderer : public VBuffer {
 
     bool show_hidden_npcs = false;
     bool boss_colors = true;
+    bool agent_border = true;
+    int agent_border_thickness;
```
DubbleClick, line 30: unsigned int agent_border_thickness = 20;

DubbleClick (changes requested):

GWToolboxdll/Widgets/Minimap/AgentRenderer.cpp
DubbleClick: (yes I realize that size_boss and size_minion don't follow this pattern, but it's predominantly used elsewhere in the project and will prevent issues if LoadSettings is ever not called)

GWToolboxdll/Widgets/Minimap/AgentRenderer.h
BearsMan: shouldn't it be int agent_border_thickness = 0; this way the user can adjust the border size to how they like it ?

DubbleClick: the user can change it one way or the other, the property should be unsigned (indicate that it's not-negative) and 20 is the default bay6 used in LoadSettings

BearsMan: that is a good point

PR #702, opened 2022-06-10
add option to flash GW window on guild message

DubbleClick:


GWToolboxdll/GWToolbox.cpp
```diff
@@ -410,7 +410,7 @@ void GWToolbox::FlashWindow() {
     flashInfo.cbSize = sizeof(FLASHWINFO);
     flashInfo.hwnd = GW::MemoryMgr::GetGWWindowHandle();
     flashInfo.dwFlags = FLASHW_TIMER | FLASHW_TRAY | FLASHW_TIMERNOFG;
-    flashInfo.uCount = 0;
+    flashInfo.uCount = 5;
```
DubbleClick, line 413: @3vcloud let me know if we need a config option for this, I think 5x is fine

PR #711, opened 2022-06-12
block packets from players on our ignore list

DubbleClick:
this is really weird as most of the time the messages are never even received by the hook, as if GW filters them out, but then sometimes they are received and TB has to block them

couldn't figure out a pattern though and this will make sure everything is blocked

@3vcloud please update the comment in GWCA saying 0 = friend, 1 = ignore, etc. It's 1 = friend, 2 = ignore, etc. like listed in the enum


DubbleClick:
Also, do we have nothing better than comparing strings for this? For local messages we could use the ID I guess but guild/ally messages don't have such a thing, do they?

3vcloud:
> Also, do we have nothing better than comparing strings for this?

Sadly not, will look at GWCA change

PR #715, opened 2022-06-25
Add Skill Monitor Widget

RiftLurker:
This is the first useable version of the skill monitor. I'll probably add a few more features later on but want to use it as is for a bit first.


3vcloud:
Nice one, will take a look in next few days

PR #718, opened 2022-06-30
using allegiance enum instead of magic numbers

franneck94:
Now using the allegiance enum of GWCA instead of magic numbers


3vcloud:
Probably better to change GWCA to declare `allegiance` as a type of `GW::Constants::Allegiance` then change the enum to uint8

Issue #725, opened 2022-07-19
Objective timer set completion time is shorter than longest objective


DubbleClick:
Can't really have the UW objectiveset finished at 24:44 when dhuum only finished at 24:46.

This happens every run.

![image](https://user-images.githubusercontent.com/7896469/179707811-6cb9340d-3a00-4a3d-8a5b-3512783c37f1.png)

PR #726, opened 2022-07-19
fix objectiveset completion time #725

DubbleClick:


3vcloud:
Stop messing with const and auto and just change what’s needed

DubbleClick:
removed all code quality changes

PR #732, opened 2022-07-25
remove intance_start from objectiveset

DubbleClick:
I don't know why this was even ever used

### DIFF
--- a/GWToolboxdll/Widgets/TimerWidget.cpp
+++ b/GWToolboxdll/Widgets/TimerWidget.cpp
@@ -280,12 +280,13 @@ std::chrono::milliseconds TimerWidget::GetRunTimeElapsed()
         return duration_cast<milliseconds>(now() - run_started);
     }
 }
-unsigned long TimerWidget::GetRunTimeStart() const
+unsigned long TimerWidget::GetStartPoint() const
 {
-    if (!is_valid(run_started)) {
+    const auto time_point = use_instance_timer ? instance_started : run_started;
+    if (!is_valid(time_point)) {
         return static_cast<unsigned long>(-1);
     } 
-    return static_cast<unsigned long>(duration_cast<milliseconds>(run_started.time_since_epoch()).count());
+    return static_cast<unsigned long>(duration_cast<milliseconds>(time_point.time_since_epoch()).count());
 }
 
 unsigned long TimerWidget::GetTimerMs() { return static_cast<unsigned long>(GetTimer().count()); }
--- a/GWToolboxdll/Widgets/TimerWidget.h
+++ b/GWToolboxdll/Widgets/TimerWidget.h
@@ -45,7 +45,7 @@ public:
     unsigned long GetTimerMs(); // time in milliseconds
     unsigned long GetMapTimeElapsedMs();
     unsigned long GetRunTimeElapsedMs();
-    unsigned long GetRunTimeStart() const;
+    unsigned long GetStartPoint() const;
     void PrintTimer(); // prints current timer to chat
 
     void OnPreGameSrvTransfer(GW::HookStatus*, GW::Packet::StoC::GameSrvTransfer* pak);
--- a/GWToolboxdll/Windows/ObjectiveTimerWindow.cpp
+++ b/GWToolboxdll/Windows/ObjectiveTimerWindow.cpp
@@ -1059,7 +1059,7 @@ ObjectiveTimerWindow::Objective* ObjectiveTimerWindow::Objective::SetStarted()
     if (IsStarted())
         return this;
     start_time_point = time_point_ms(); // run_started_time_point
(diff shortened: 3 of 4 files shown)


3vcloud:
A run is started when you've loaded in; you might load in slower than other people so objective 1 could start late. I think this would throw the whole run time off a bit if you loaded in late.

3vcloud:
Yeh looks like this now calculates everything from run start i.e. when the map loaded for your player - this is wrong and it needs to be against the instance time, hence the calculation for `done`

DubbleClick:
yeah that's why I came to it in the first place, we shouldn't be using 3 different timers but only one that is actually relevant

DubbleClick:
(run_time_point_started already uses the TimerWidgets time now btw)

3vcloud:
There will almost never be a situation when the run start time point is equal to the instance start time point unless you're plugged into the datacenter. The objective timer has been combed over on this one and I think the timestamps will be messed up if you use this logic now.

DubbleClick:
wouldn't that mean that the instancetimer/timerwidget is wrong?

3vcloud:
The timer widget should start its instance timer from the moment the map starts loading, which is a deliberate change made around a year ago because the GW server's instance time is not always as accurate (there is a feature in the timer widget to show both of these timestamps on screen to show the difference). This timestamp should be treated as the instance time, and any other timestamps recorded within the instance based off of it.

DubbleClick:
I'm happy to switch back to instance time for everything by the way. I just don't want 3 different timers ticking when only one is useful for the end completion time and another is factually "correct"

DubbleClick:
> This timestamp should be treated as the instance time, and any other timestamps recorded within the instance based off of it.

In that case this should just be fine now, no?

3vcloud:
No, times are now going to be relative to when the ObjectiveSet was created rather then when the map was created, which in effect depends more on individual player load times than the team's achievement in the map

DubbleClick:
So what's the solution where we use *one* time instead of three

DubbleClick:
> The timer widget should start its instance timer from the moment the map starts loading

To make it clear again, this is also the time that the run_start_time_point from the ObjectiveSet uses

3vcloud:
No, it uses `TimerWidget::Instance().GetRunTimeStart()` which is different to `TimerWidget::Instance().GetTimeElapsedMs()`

DubbleClick:
`TimerWidget::Instance().GetTimeElapsedMs()` uses that very time point for its calculation

3vcloud:
```cpp
std::chrono::milliseconds TimerWidget::GetMapTimeElapsed() {
    if (!is_valid(instance_started)) {
        // Can happen if toolbox was started after map load
        instance_started = now() - milliseconds(GW::Map::GetInstanceTime());
    }
    return duration_cast<milliseconds>(now() - instance_started);
}
...
unsigned long TimerWidget::GetRunTimeStart() const
{
    if (!is_valid(run_started)) {
        return static_cast<unsigned long>(-1);
    } 
    return static_cast<unsigned long>(duration_cast<milliseconds>(run_started.time_since_epoch()).count());
}
...
unsigned long TimerWidget::GetMapTimeElapsedMs() { return static_cast<unsigned long>(GetMapTimeElapsed().count()); }
```
`GetRunTimeStart()` is not the same as `GetMapTimeElapsedMs()`

DubbleClick:
```cpp
std::chrono::milliseconds TimerWidget::GetTimer() {
    if (use_instance_timer) {
        return milliseconds(instance_timer_valid ? GW::Map::GetInstanceTime() : 0);
    }
    return GetRunTimeElapsed();
}
```

I've updated the code to check use_instance_timer

3vcloud:
1. The timer widget is supposed to show you on-screen how long the instance has been up.
2. All objectives are relative to when the instance was started, not the run. This is because players load in at different times and their client would start the objective sets later if they loaded in later.
3. Some runs span 1 or more maps, and so the run time could be greater than the instance time.

Is there something not working right with the objective timer that needs all of these changes?

DubbleClick:
> Is there something not working right with the objective timer that needs all of these changes?

As things were, the ObjectiveSet uses a different time than the TimerWidget and the Objective completions use yet another time. That's three different times when only one is actually relevant to us (either the instance time or the run time used for gwscr).

DubbleClick:
> `GetRunTimeStart()` is not the same as `GetMapTimeElapsedMs()`

TimerWidget::GetMapTimeElapsed() has zero references throughout the project, by the way. It's unused

3vcloud:
I'm just going to merge this in, we can test and fix later if needed

DubbleClick:
I've already been testing it for what it's worth and it's now working as expected  - ObjectiveSet, Objective and TimerWidget use the same time points.

PR #733, opened 2022-07-27
add option to not flip minimap on reverse camera (default)

DubbleClick:
this has been bugging me for a while now, the regular gw compass doesn't flip so the mismatch was irritating

I've made it the default behaviour that the minimap does not flip on reverse


DubbleClick:
this also fixes the Reapply Title behaviour now if Title::title_id gets fixed too

3vcloud:
Minimap changes merged

PR #745, opened 2022-08-09
changes

DubbleClick:
I'm at a complete loss why it says I've made changes in GWCA - I haven't. rebasing and making sure the files match the exact versions of you latest commit reveal no differences either.


GWToolboxdll/Modules/DialogModule.cpp
```diff
@@ -175,6 +174,31 @@ const std::vector<GuiUtils::EncString*>& DialogModule::GetDialogButtonMessages()
...
+    if (dialog_buttons.empty()) return 0;
+    const GW::UI::DialogButtonInfo* to_use = nullptr;
+    for (const auto dialog_button : dialog_buttons) {
+        switch (dialog_button->button_icon) {
+            case 16: // quest accept icon
+            case 18: // quest enquire icon
+            case 23: // quest reward icon
+                to_use = dialog_button;
+        }
+        if (to_use && to_use->dialog_id == 0x806703) continue; // skip uwg unless it's the only available dialog
```
3vcloud, line 188: Can we do a bitwise XOR on this against the quest ID for UWG to make it readable without comments?

GWToolboxdll/Windows/Hotkeys.cpp
```diff
@@ -1614,15 +1612,22 @@ void HotkeyDialog::Execute()
...
+            }
+        } else {
+            DialogModule::AcceptFirstAvailableQuest();
+        }
         return;
+    }
     if (isExplorable()) {
-        GW::Agent *target = GW::Agents::GetTarget();
-        if (target && target->type == 0xDB) {
+        if (const auto* target = GW::Agents::GetTarget(); target && target->type == 0xDB) {
```
3vcloud, line 1626: AsAgentLiving mebe

GWToolboxdll/Modules/DialogModule.cpp
```diff
@@ -97,11 +97,10 @@ namespace {
         }
     }
     bool IsDialogButtonAvailable(uint32_t dialog_id) {
-        for (auto d : dialog_buttons) {
-            if (d->dialog_id == dialog_id)
-                return true;
-        }
-        return false;
+        return std::any_of(
```
3vcloud, line 100: Just to check, `any_of` does drop out after the first resolve right?

DubbleClick: we can only make educated guesses there as the actual implementation is compiler dependent, but it's fair to assume that the STL method would be at least as efficient as a manual loop

DubbleClick:
Actual implementation of std::any_of in MSVC v142 with C++17:

```
template <class _InIt, class _Pr>
_NODISCARD _CONSTEXPR20 bool any_of(const _InIt _First, const _InIt _Last, _Pr _Pred) {
    // test if any element satisfies _Pred
    _Adl_verify_range(_First, _Last);
    auto _UFirst      = _Get_unwrapped(_First);
    const auto _ULast = _Get_unwrapped(_Last);
    for (; _UFirst != _ULast; ++_UFirst) {
        if (_Pred(*_UFirst)) {
            return true;
        }
    }

    return false;
}
```

PR #755, opened 2022-08-12
last QoL changes

DubbleClick:


GWToolboxdll/Modules/GameSettings.cpp
```diff
@@ -140,27 +139,27 @@ namespace {
...
-    static bool ctrl_enter_whisper = false;
+    bool ctrl_enter_whisper = false;
 
-    static bool IsInfused(GW::Item* item) {
+    bool IsInfused(GW::Item* item) {
         return item && item->info_string && wcschr(item->info_string, 0xAC9);
     }
 
-    const enum PING_PARTS {
+    enum PING_PARTS {
```
DubbleClick, line 162: this is a funny one, never heard of a const enum before and the standard doesn't define it. interesting that it even compiles in MSVC, in gcc this is a compiler error

PR #757, opened 2022-08-12
prevent unwanted copies by looping with const references

DubbleClick:
noticed we had a lot of these, partially consuming a lot of extra memory (plenty of std::pairs or even std::vectors), this doesn't change anything but to loop via (const) reference for performance and memory sake

may not have caught all of them (only searched for `for (auto ` and fixed manually

### DIFF
--- a/GWToolboxdll/Widgets/Minimap/EffectRenderer.cpp
+++ b/GWToolboxdll/Widgets/Minimap/EffectRenderer.cpp
@@ -56,36 +56,36 @@ void EffectRenderer::LoadDefaults() {
     aoe_effect_triggers.emplace(Spike_Trap_Activate, new EffectTrigger(Spike_Trap, 2000, GW::Constants::Range::Nearby));
 }
 EffectRenderer::~EffectRenderer() {
-    for (auto settings : aoe_effect_settings) {
+    for (const auto& settings : aoe_effect_settings) {
         delete settings.second;
     }
     aoe_effect_settings.clear();
-    for (auto triggers : aoe_effect_triggers) {
+    for (const auto& triggers : aoe_effect_triggers) {
         delete triggers.second;
     }
     aoe_effect_triggers.clear();
 }
 void EffectRenderer::Invalidate() {
     VBuffer::Invalidate();
-    for (auto p : aoe_effects) {
+    for (const auto p : aoe_effects) {
         delete p;
     }
     aoe_effects.clear();
-    for (auto trigger : aoe_effect_triggers) {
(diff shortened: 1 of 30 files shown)


3vcloud:
Aren’t half of these pointers/uints anyway, so no gain?

DubbleClick:
No, I've not made a pointer a reference (only added const in a few of those cases). Every reference added is a data struct, typically a pair, once or twice a SkillActivation struct and once even a vector.

Iterating with const ref pointers would be 'pointless' ;)

3vcloud:
Hoho good one! But aren’t iterators kind of pointers, you’ve just changed what the loop uses to iterate now which has the same meaning?

DubbleClick:
Those aren't iterators though, they're std::pairs.

I don't know why half of these are called `it`

GWToolboxdll/Widgets/Minimap/EffectRenderer.cpp
```diff
@@ -56,36 +56,36 @@ void EffectRenderer::LoadDefaults() {
...
-    for (auto triggers : aoe_effect_triggers) {
+    for (const auto& triggers : aoe_effect_triggers) {
         delete triggers.second;
     }
     aoe_effect_triggers.clear();
 }
 void EffectRenderer::Invalidate() {
     VBuffer::Invalidate();
-    for (auto p : aoe_effects) {
+    for (const auto p : aoe_effects) {
```
DubbleClick, line 70: example of a raw pointer, no & added

GWToolboxdll/Widgets/Minimap/EffectRenderer.cpp
```diff
@@ -56,36 +56,36 @@ void EffectRenderer::LoadDefaults() {
     aoe_effect_triggers.emplace(Spike_Trap_Activate, new EffectTrigger(Spike_Trap, 2000, GW::Constants::Range::Nearby));
 }
 EffectRenderer::~EffectRenderer() {
-    for (auto settings : aoe_effect_settings) {
+    for (const auto& settings : aoe_effect_settings) {
         delete settings.second;
     }
     aoe_effect_settings.clear();
-    for (auto triggers : aoe_effect_triggers) {
+    for (const auto& triggers : aoe_effect_triggers) {
```
DubbleClick, line 63: example of std::pair<int, pointer>

DubbleClick (commented):
can't find the std::vector atm, may have even already changed that in the other pr (which made me do a very rough search for loops here)

3vcloud:
Ok thats incredibly stupid of me, nice catch

PR #759, opened 2022-08-13
Cpp20

DubbleClick:


DubbleClick:
it looks massive, but it doesn't contain any functionality changes

the three big things are:

icons are now char8_t*, but can be safely casted to const char*
const char[] literals can not be assigned to char* (technically UB since c++11)
c2362 (variable declaration skipped by goto statements) cannot be ignored anymore

the last 4 commits re-enable warnings that we don't have to supress anymore and switch container based loops to use std::ranges

GWToolboxdll/Modules/ChatCommands.cpp
```diff
@@ -1594,43 +1594,40 @@ void ChatCommands::CmdTransmoTarget(const wchar_t*, int argc, LPWSTR* argv) {
...
+        memset(&hom_achievements, 0, sizeof(hom_achievements));
+        HallOfMonumentsModule::Instance().AsyncGetAccountAchievements(
+            player_name, &hom_achievements, &hom_achievements_result);
+    } else {
+        Log::Error("Invalid player name for hall of monuments command");
+    }
+}
+
+void ChatCommands::CmdHom(const wchar_t* message, int argc, LPWSTR*) {
+    wchar_t player_name[20]{};
```
3vcloud, line 1608: Deliberately avoided keeping it on the stack because it’s hardly ever used

DubbleClick: I personally feel like the 20 bytes are preferable to having to do manual memory management, even if it's contained within the same function

happy to change it back if you insist though

3vcloud: Nvm this is probably a bad example of me using new anyway

PR #760, opened 2022-08-13
more references, gwca update, dialogmodule addition

DubbleClick:


3vcloud:
Merged, but stop making `consts` and `autos` where they're not needed

PR #765, opened 2022-08-15
Dev

DubbleClick:


GWToolboxdll/imgui_impl_win32.cpp
```diff
@@ -158,6 +158,11 @@ void    ImGui_ImplWin32_NewFrame()
     io.KeySuper = false;
     // io.KeysDown[], io.MousePos, io.MouseDown[], io.MouseWheel: filled by the WndProc handler below.
 
+    // dirty hack to disable imgui keyboard navigation
+    // TODO: remove once imgui lets us disable keyboard navigation entirely
+    // see: https://github.com/ocornut/imgui/issues/4828
+    io.KeyCtrl &= (::GetKeyState(VK_TAB) & 0x8000) == 0;
+
```
DubbleClick, line 165: really not happy with this, but it works and leaves other tabbing in imgui intact, which is important

should definitely remove this once the NavEnableKeyboard config flag is checked

GWToolboxdll/Modules/DialogModule.cpp
```diff
@@ -139,8 +121,58 @@ namespace {
...
+            case QuestDialogType::REWARD: break;
+            default: return;
+        }
+        for (auto it = queued_dialogs_to_send.begin(); it != queued_dialogs_to_send.end();) {
+            const auto other_dialog_id = it->first;
+            // make sure we don't delete dialogs for the same quest queued up later, e.g. separate reward dialog!
+            if (GetQuestID(other_dialog_id) == quest_id && queued_at >= it->second) {
+                it = queued_dialogs_to_send.erase(it);
+            } else {
+                it++;
```
DubbleClick, line 168: so this whole logic change is actually important, if we didn't have this check, separate take/reward dialog ids would be cleared and only the take action would be executed

GWToolboxdll/Modules/DialogModule.cpp
```diff
@@ -162,56 +194,58 @@ void DialogModule::Initialize() {
         GW::HookBase::EnableHooks(NPCDialogUICallback_Func);
     }
 }
-void DialogModule::SendDialog(uint32_t dialog_id) {
-    const bool already_queued = queued_dialogs_to_send.contains(dialog_id);
-    queued_dialogs_to_send[dialog_id] = TIMER_INIT();
-    if (already_queued)
-        return; // Don't redo any logic.
```
DubbleClick, line 169: had to remove this because the new dialog trigger should refresh all related dialogs as well, otherwise nothing might end up happening (previous dialog missing) even though the hotkey was last pressed <3s ago.

GWToolboxdll/Windows/DialogsWindow.cpp
```diff
@@ -119,16 +119,15 @@ void DialogsWindow::Draw(IDirect3DDevice9* pDevice) {
...
-                    GoNPCSendDialog(QuestAcceptDialog(IndexToQuestID(fav_index[i])));
-                }
-                ImGui::SameLine(0, ImGui::GetStyle().ItemInnerSpacing.x);
-                if (ImGui::Button("Reward", ImVec2(60.0f, 0))) {
-                    GoNPCSendDialog(QuestRewardDialog(IndexToQuestID(fav_index[i])));
+                if (ImGui::Button("Send", ImVec2(60.0f, 0))) {
+                    GoNPCSendDialogs({
+                        QuestAcceptDialog(IndexToQuestID(fav_index[i])),
+                        QuestRewardDialog(IndexToQuestID(fav_index[i]))
+                    });
```
DubbleClick, line 130: could implement `double click = queue both separately` in the future meaning that take and reward are both executed
not really too important for now I feel

DubbleClick (commented):
some clarifications for the rationale behind the logic changes

DubbleClick:
rebased to last dev

PR #770, opened 2022-08-19
Added hide items at merchant

franneck94:
Added option of item ids to hide in merchant window.
Like hiding your lockpicks to don't sell them by accident

### DIFF
--- a/GWToolboxdll/Modules/InventoryManager.cpp
+++ b/GWToolboxdll/Modules/InventoryManager.cpp
@@ -672,8 +672,8 @@ void InventoryManager::Initialize() {
 // Hide unsellable items from merchant
 void InventoryManager::OnAddItemToWindow(void* ecx, void* edx, uint32_t frame, uint32_t item_id) {
     GW::Hook::EnterHook();
-    GW::Item* item = Instance().hide_unsellable_items ? GW::Items::GetItemById(item_id) : 0;
-    if (!item || item->value)
+    GW::Item* item = (Instance().hide_unsellable_items || Instance().hide_items.size()) ? GW::Items::GetItemById(item_id) : 0;
+    if (!item || (item->value && std::find(Instance().hide_items.begin(), Instance().hide_items.end(), item->model_id) == Instance().hide_items.end()))
         RetAddItemRowToWindow(ecx, edx, frame, item_id);
     GW::Hook::LeaveHook();
 }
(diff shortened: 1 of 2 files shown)


GWToolboxdll/Modules/InventoryManager.cpp
```diff
@@ -744,6 +744,11 @@ void InventoryManager::SaveSettings(CSimpleIni* ini) {
     ini->SetBoolValue(Name(), VAR_NAME(wiki_link_on_context_menu), wiki_link_on_context_menu);
     ini->SetBoolValue(Name(), VAR_NAME(hide_unsellable_items), hide_unsellable_items);
     ini->SetBoolValue(Name(), VAR_NAME(change_secondary_for_tome), change_secondary_for_tome);
+
+    const char* const section = "hide_items";
```
3vcloud, line 748: Keep module settings to their own section

GWToolboxdll/Modules/InventoryManager.cpp
```diff
@@ -1322,6 +1332,31 @@ void InventoryManager::DrawSettingInternal() {
...
+    }
+    ImGui::EndChild();
+    ImGui::Separator();
+    bool submitted = false;
+    ImGui::Text("Add new item from merchant window:");
+    ImGui::InputInt("Item ID", &new_item_id);
+    submitted |= ImGui::Button("Add");
+    if (submitted) {
+        AddItemID(new_item_id);
+        Log::Info("Added Iten ID (%d)", new_item_id);
```
3vcloud, line 1358: typo

GWToolboxdll/Modules/InventoryManager.h
```diff
@@ -1,5 +1,7 @@
 #pragma once
 
+#include <vector>
```
3vcloud, line 3: Instead of having to include `<vector>` in the header, define this in the private namespace of `InventoryManager.cpp`, along with the function definitions for `AddItemID` and `RemoveItemID`

GWToolboxdll/Modules/InventoryManager.h
```diff
@@ -94,7 +96,16 @@ class InventoryManager : public ToolboxUIElement {
     bool salvage_rare_mats = false;
     bool show_transact_quantity_popup = false;
     bool transaction_listeners_attached = false;
+
     bool hide_unsellable_items = true;
+    std::vector<uint32_t> hide_items = {};
+    int new_item_id = 0;
+    void AddItemID(const uint32_t item_id) {
```
3vcloud, line 103: Probably want to rename these to some thing like `AddMerchantHiddenItemID` - this module is used for a lot of item based stuff and this function could get a bit confusing later

GWToolboxdll/Modules/InventoryManager.h
franneck94: But we have a member variable that is a vector declared in the header, so why not include the vector header here?

PR #772, opened 2022-08-21
Dev

DubbleClick:


GWToolboxdll/Windows/Hotkeys.cpp
```diff
@@ -1511,12 +1511,10 @@ void HotkeyTarget::Execute()
         break;
     default:
         Log::ErrorW(L"Unknown target type %d", type);
-        delete[] message;
         return;
     }
     GW::GameThread::Enqueue([message]() {
         GW::Chat::SendChat('/', message);
-        delete[] message;
```
3vcloud, line 1519: This is why its necessary

DubbleClick: copies message by value with default capture

3vcloud: Right you are, but why copy?

DubbleClick: compiler optimises it away into a move anyway, pretty sure

memory safety > minimal performance gain (though with optimisations most likely a performance penalty for the heal allocation anyway)

PR #793, opened 2022-09-16
update openssl (no more lnk4098 in release mode)

DubbleClick:
in theory we should also build libcurl and openssl with /MTd and use that for debug compilations, but I can live with the warning in debug mode


DubbleClick:
relevant code changes:
https://github.com/HasKha/GWToolboxpp/pull/793/files#diff-3ecb8e84ada2c4bc7c927bf6e705aee4f16b006aa31397847abb84e0e2001d42R3

PR #794, opened 2022-09-16
use wolfSSL instead of OpenSSL

DubbleClick:
this saves 2.8mb of file size in RelWithDbgInfo (6.9mb instead of 9.7mb)


Dependencies/easywsclient/easywsclient.cpp
```diff
@@ -522,6 +522,7 @@ easywsclient::WebSocket::pointer from_url(const std::string& url, bool useMask,
           ERR_print_errors_cb(print_error_callback,0);
 
       // Create an SSL struct for the connection
+      SSL_CTX_set_verify(ptConnCtx->sslContext, SSL_VERIFY_NONE, NULL);
```
DubbleClick, line 525: this is crucial, wolfSSL verifies the servers certificate with loaded CA's by default,. while openSSL doesn't

DubbleClick:
down to 4.2mb in relwithdebinfo configuration

PR #817, opened 2022-09-30
various things

DubbleClick:


GWToolboxdll/Modules/GameSettings.cpp
```diff
@@ -476,11 +476,12 @@ namespace {
             uint32_t source;
             uint32_t identifier;
         } *party_target_info = (Packet*)wparam;
-        current_party_target_id = 0;
         switch (message_id) {
         case GW::UI::UIMessage::kTargetPlayerPartyMember: {
-            if (!IsPlayerInParty(party_target_info->identifier))
```
3vcloud, line 482: Be sure to check this works when targeting an invitee from party window

DubbleClick: it doesn't work for people you invite or who you were invited by, I'll implement that tomorrow or so when I have time
party context contains the invites so I can check there

3vcloud: This isn’t what I meant, there is no point re inviting someone who is in your party yet anyway. What does the game do if you target someone who isn’t in your party? Doesn’t it keep the previous party member highlighted?

3vcloud:
Heads up, want to change the way obfuscator works to avoid renaming player names in the stoc packet but haven’t found a good place to hook the party window yet so don’t try to change too much with the logic

DubbleClick:
Ah thanks, it still doesn't obfuscate on start up but I'll let it be then.

PR #835, opened 2022-10-30
Improve autoclicker performance

tjubutsi:
Not sure if this will give issues for FPS under 144, couldn't be arsed to test, but I expect it won't.


3vcloud:
Have you changed the setting to allow it to go this low?

tjubutsi:
Is that a thing? It worked out of the box with just this change for me. There's no settings for the autoclicker (afaik).

3vcloud:
in dev ta

PR #862, opened 2023-01-10
feat: add duping window

MikeJerred:
Added a window to help with DoA duping

### DIFF
--- a/GWToolboxdll/Modules/ToolboxSettings.cpp
+++ b/GWToolboxdll/Modules/ToolboxSettings.cpp
@@ -56,6 +56,7 @@
 #include <Windows/ObserverPartyWindow.h>
 #include <Windows/ObserverExportWindow.h>
 #include <Windows/CompletionWindow.h>
+#include <Windows/DupingWindow.h>
 #ifdef _DEBUG
 #include <Windows/PacketLoggerWindow.h>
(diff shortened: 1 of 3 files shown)


BearsMan:
Nice job Mike! This should be in Dev branch not Master, but I am sure it will be worked out soon.

MikeJerred:
@BearsMan thanks, I switched base and there are no conflicts so should be fine

BearsMan:
> @BearsMan thanks, I switched base and there are no conflicts so should be fine

Yea, as long as you see the words no conflicts with your code. Then that means it will most likely be put into the master branch after it has been tested in dev branch

BearsMan:
Yea, as long as you see the words no conflicts with your code. Then that means it will most likely be put into the master branch after it has been tested in dev branch

3vcloud:
See https://github.com/HasKha/GWToolboxpp/commit/56b647b986f0a936f37bfab8a3dc63ef4cc43c9c for post merge changes

3vcloud:
Its worth also noting that this will only work properly when GW is the focussed window; the Draw() and Update() calls don't run every frame if the game isn't being played, so need to think about hooking StoC packets to maintain the vectors outside of the draw loop

PR #892, opened 2023-04-21
Fix colour of NPC allies on minimap

EnergyFromGW:
It would appear that the colour used for "Non-attackable allies" is always `color_ally`, corresponding to the "Ally (player)" setting. There are however two types of agents that would fall under this category: NPCs which are henchmen/heroes (which should use `color_ally_npc`), and player allies.

The issue can be seen by going into an outpost with henchmen and observing that the colour used is the one intended for player allies.

The fix is to check whether the agent is an NPC before deciding which of two colours to use.

Attached are a couple of images to demonstrate.

Tested: Outpost + Explorable.

energy#4904

*Minimap as of master:*, Sunspear Great Hall, observe how henchmen use the player ally colour.
original

*Minimap as of this commit:*
fixed

*Settings used:*
settings

### DIFF
--- a/GWToolboxdll/Widgets/Minimap/AgentRenderer.cpp
+++ b/GWToolboxdll/Widgets/Minimap/AgentRenderer.cpp
@@ -887,7 +887,7 @@ Color AgentRenderer::GetColor(const GW::Agent* agent, const CustomAgent* ca) con
     // friendly
     if (living->GetIsDead()) return color_ally_dead;
     switch (living->allegiance) {
-    case GW::Constants::Allegiance::Ally_NonAttackable: return color_ally; // ally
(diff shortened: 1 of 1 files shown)


EnergyFromGW:
Note that this does also change the minimap colour for henchmen to the NPC ally colour in explorables -- which although is technically correct, it might be unexpected from users. If that's not desirable, you could also check whether you're in an outpost before returning `color_ally_npc`.

3vcloud:
Hey up, so this one is a toss up between what players are used to seeing with toolbox vs what should be shown in game. We want to impact existing settings as little as possible on this one, and changing the colour of allied henchmen in explorable is a big no for this reason.

having said that, it’s much less invasive on gameplay if we changed the colour of henchmen in an outpost to the npc colour, but then you need to think is it worth an explorable area check just for this tweak.

Tldr don’t add another setting, but possibly force henchmen in outpost to read from npc colour instead

Issue #842, opened 2022-12-16
flash taskbar when everyone else is ticked


Hour-of-the-Owl:
If you're in a team with at least one other human, the GW icon on your taskbar should flash when everyone else has ticked up. Regardless of whether you yourself have ticked, this is a very good indicator that you should tab back into GW because your team will be waiting for you.
I guess this should be optional, but I can't really think of a use case where it would make sense to turn it off, so at least have it on by default.

PR #893, opened 2023-04-21
 #842: flash when all other party members ready

EnergyFromGW:
Implements #842 

energy#4904


DubbleClick:
thank you

DubbleClick:
actually this existed in ToastNotifications, but I can see how that's confused and will rework them all to be in the same module

PR #911, opened 2023-06-29
Add default values for minimap in default position

adituv:
If the compass UI component has never been moved or resized from its default position and size, the minimap widget would not display, due to the default bounds being returned as all zeroes.  This patch adds some default values for that edge case to work around the missing values in the compass frame's bounds.


adituv:
Oh my bad; do you need me to close this and re-submit to dev?

3vcloud:
Nope, I’ll pick it into dev and close this off

PR #916, opened 2023-08-03
Fix damage widget offset with hard mode button shown

adituv:
Move the damage widget down when in an outpost with the NM/HM buttons showing, so that the meters are in line with the health bars.


adituv:
Apologies for posting against the wrong branch.  I'll do a new one.

Issue #816, opened 2022-09-30
/reinvite not working


Hour-of-the-Owl:
`/reinvite` doesn't seem to work - just sends the error message "Choose target party member to reinvite"

DubbleClick:
[7646315](https://github.com/HasKha/GWToolboxpp/pull/817/commits/7646315c6f170285eb3d49d7c850c24a1e3fa962)

#817

3vcloud:
https://github.com/HasKha/GWToolboxpp/commit/0243a630f1fab002524719fb83ddb098314eb0fb

Hour-of-the-Owl:
1. I don't remember which TB version broke it, but it used to be possible to invite people who aren't in your party at all. This no longer works, and doesn't even give an error message.
2. When targeting some people, the feature works correctly, but for most I just get the `choose a party member to reinvite` error.

3vcloud:
1. Has never worked this way, and for a good reason - the yellow highlight on the party window is not directly related to the in game targeting system and there are some situations where the person highlighted in the party window doesn’t match the current target window in game.
2. Just to check, are the people you’re targeting to reinvite already party members or not, and are you targeting them in game or via the party window, related to the point above?

Hour-of-the-Owl:
1. Unless I have Alzheimer's, I distinctly remember it working that way, but it sounds like you're misunderstanding me.
What I want to be able to do is to target anyone in the outpost (including henchmen), use `/reinvite`, and have it send a party invitation to them. If they were already in the party and I'm the leader, it should kick them before reinviting. If I am not the party leader, it should first leave the party - the exception would be if I currently have a hero selected.
2. For this point, the people I'm targeting are already in the party, and I'm the leader, trying to reorder them. It makes no difference which way I target them. Right now I'm in a party of 4 of my own accounts, and `/reinvite` works correctly on just one member, regardless of how I selected them. When I put a different account as leader, it has the same problem, but with a different party member. Most of the time, targeting myself leaves and reinvites, but not always.

adituv:
Following on from #918, it appears that there are multiple "parties" that make up an overall party, particularly when heroes are involved.  The parties are not necessarily the same between players.  On one account, my character with a hero has a different PartyInfo to the character without a hero - the former includes heroes and the latter does not; I assume to handle inviting and dismissing heroes independently of other players.

[...]

PR #918, opened 2023-08-04
Fix reinvite command

adituv:
See #816

IsPlayerInParty was not consistently returning accurate results, and this had a knock-on effect of breaking the /reinvite command.  Now it is much simpler conceptually and /reinvite appears to work.

### DIFF
--- a/GWToolboxdll/Utils/ToolboxUtils.cpp
+++ b/GWToolboxdll/Utils/ToolboxUtils.cpp
@@ -345,8 +345,15 @@ namespace ToolboxUtils {
 
     bool IsPlayerInParty(const uint32_t login_number)
     {
-        GW::PartyInfo* party = nullptr;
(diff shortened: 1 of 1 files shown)


3vcloud:
Looks like you've sidestepped a potential bug with a function higher up the chain rather than fixing it with this one? Surely there must be an issue with `GetPlayerPartyMember` in this case?

3vcloud:
I would instead roll this back, break on the timeout reached [in `UpdateReinvite`](https://github.com/HasKha/GWToolboxpp/blob/547d05821e8938664be170c6fbe3915a8cf5b04f/GWToolboxdll/Modules/GameSettings.cpp#L561) and take a look at the situation at this time - your fix technically is no different to the existing code, and we need to find out why its not working consistently to fix the problem

PR #921, opened 2023-08-09
add ability to draw custom markers in game (on terrain)

xenotrace:
this PR adds the ability to draw custom TB minimap markers in-game (on the terrain).

there is a checkbox option for each custom line/marker/polygon (default `false`) to control whether it is draw this way.
there are two additional settings specific for this code:
image


DubbleClick:
pull requests please always onto `dev` branch, I will pick this into dev

PR #924, opened 2023-08-10
6.11

DubbleClick:
* [New] Added item upgrade unlocks to completion window
* [New] Added chat filter for dropped ashes (@adituv)
* [New] Added GameWorldRenderer to render lines, circles and polygon on the game terrain. Note that this isn't fully fleshed out yet and draws over your character, props and walls. (@xenotrace) 
* [Minor] Added default values for the Minimap position if you've never moved the in-game compass
* [Minor] Toolbox widgets now all remember their 
* [Minor] Added option to auto use keys on chests (prioritised before lockpicks)
* [Minor] Added option to get notified when resurrected
* [Minor] Damage monitor, bonds widget and cast widget now account for the different party member height in outposts
* [Fix] Fixed option to auto-login when no character name parameter is passed to gw. Requires you to launch with GW Launcher
* [Fix] Removed copy hero build button when no builds existed.
* [Fix] Fixed bug preventing secondary profession switch on tome use.

we should prepare a release soon


BearsMan:
I noticed in the notes that the section for Toolbox Widgets isn't fully written in a complete sentence.

DubbleClick:
thanks, fixed

BearsMan (approved):

BearsMan:
> Thanks, fixed

No worries!

.clang-format
```diff
@@ -12,25 +13,25 @@ AllowShortIfStatementsOnASingleLine: true
 AllowShortLambdasOnASingleLine: false
 AlwaysBreakTemplateDeclarations: Yes
 BraceWrapping: 
-  AfterCaseLabel: true
-  AfterClass: true
+  AfterCaseLabel: false
```
franneck94, line 16: this braces style makes me sad :(

DubbleClick: many of these styles aren't my personal favourite, but I based the decisions on the majority of already written code

only made a few (very necessary) changes and decided for preference on things that were evenly split across the code base

DubbleClick: if you have input on specific things you'd like to see changed, bring them up in a pr (run clang-format and clang-tidy on a good example file) and let jon decide if he wants to merge it

GWToolboxdll/Modules/ToastNotifications.cpp
```diff
@@ -52,42 +52,42 @@ namespace {
     // Pointers deleted in ChatSettings::Terminate
     std::map<std::wstring, ToastNotifications::Toast*> toasts;
 
-    bool CanNotify()
+    bool show_toast(bool show_toast)
```
franneck94, line 55: i dont like having the same func and param name

.clang-format
```diff
@@ -1,237 +1,66 @@
...
+  BeforeLambdaBody: false
+  BeforeWhile: false
+  SplitEmptyFunction: false
+  SplitEmptyRecord: false
+  SplitEmptyNamespace: false
+BreakBeforeBraces: Custom
+BreakConstructorInitializers: BeforeColon
+BreakInheritanceList: AfterColon
+BreakStringLiterals: false
+ColumnLimit: 256
```
franneck94, line 36: Should we maybe set the line length to e.g. 120 isntead of this 256?
Otherwise we could have a very long line inbeteween of shorter lines which is bad for thr readability of the code.

franneck94: Also, is it planned to run clang-format on all files in this PR?
At some point we should (?) do this.

DubbleClick: Set it to 256 because that was auto-detected. Anything around 120-140 sounds fine to me.

All of the GWToolboxdll has been formatted already since this is where most of the code we work on is.

franneck94: I asked because i ran clang-format on the GWToolboxdll files and i had changes in 50 files.
Most changes related to the line length or sorted includes.

@3vcloud What is your opinion on the line length?

DubbleClick: I set a couple extra Resharper rules so that could explain the difference.

BearsMan:
Since the GW update broke TB, we should make a 6.10.1 hotfix instead of pushing to Version 6.11 right away since most of the features of 6.11 from what I gathered in testing are not fully finished yet.

DubbleClick:
What's broken?

BearsMan:
Disregard my previous comment about the hot fix.

BearsMan:
> What's broken?

Nothing broke, I thought that due to game build update that it would brake some signatures, but it works fine for now, I had to lookup what the update was about and it is only minor regarding an NPC (GTOB)

DubbleClick:
Okay, so nothing to do. 6.11 is ready anyhow. The thing that isn't ready yet won't be for many months, no reason not to tag a release before that.

GWToolboxdll/Modules/InventoryManager.cpp
```diff
@@ -456,94 +556,293 @@ namespace {
...
+    void __fastcall OnAddItemToWindow(void* ecx, void* edx, const uint32_t frame, const uint32_t item_id)
+    {
+        GW::Hook::EnterHook();
+        if (merchant_list_tab == 0xb && hide_unsellable_items) {
+            const auto item = GW::Items::GetItemById(item_id);
+            if (item && !item->value) {
+                GW::Hook::LeaveHook();
+                return;
+            }
+        }
```
ChandlerFerry, line 762: This code no longer handles `hide_from_merchant_items` ([as intended](https://github.com/HasKha/GWToolboxpp/blob/master/GWToolboxdll/Modules/InventoryManager.cpp#L586-L593)).

Causes the manually hidden items list to be ignored.

---

Attempted to solve this, but am not experienced enough with C++ to understand how to best access `hide_from_merchant_items` when `onAddItemToWindow` is not of the `InventoryManager` class.

Possible solutions:

1. Set `hide_from_merchant_items` similarly to how [hide_unsellable_items](https://github.com/HasKha/GWToolboxpp/blob/dev/GWToolboxdll/Modules/InventoryManager.cpp#L54) is set

2. Create a public InventoryManager function `IsHiddenMerchantItem` that returns a boolean value?

3. Something I haven't thought of (highly likely)

Would appreciate insight as to what direction to go.

PR #927, opened 2023-08-13
rename players according to friends list alias

xenotrace:
This PR adds a setting to the party window module (defaulted to **off**) to rename players according to an alias set in the friends list.

## Example:
image

## Setting location:
image

## Functionality tested:
- [x] players not on friend list are unaffected
- [x] works in explorable and outpost
- [x] friends list entry with no alias is unaffected
- [x] trading / following works on renamed players (assuming other functionality is similarly unaffected too)
- [x] changing alias doesn't break anything for current map session


3vcloud:
Make sure this works with the reinvite command

DubbleClick:
picked into dev, did some formatting before so couldn't merge here

PR #930, opened 2023-08-14
prevent rendering custom world items in outposts

xenotrace:
in DoA, outpost map ID matches explorable map ID, and the terrain is not the same.


BearsMan:
I noticed that when I tested the feature.

PR #941, opened 2023-09-28
Added missing outposts to the travel window

quetzalsly:
Works in presearing and postsearing.
If the player has unlocked everything then the combo box should not be drawn at all.
Haven't tested this in game because I don't have a char that has unlocked everything yet.

### DIFF
--- a/GWToolboxdll/Windows/TravelWindow.cpp
+++ b/GWToolboxdll/Windows/TravelWindow.cpp
@@ -54,10 +54,10 @@ namespace {
     bool IsAlreadyInOutpost(const GW::Constants::MapID outpost_id, const GW::Constants::District _district, const uint32_t _district_number = 0)
     {
         return GW::Map::GetInstanceType() == GW::Constants::InstanceType::Outpost
-               && GW::Map::GetMapID() == outpost_id
-               && TravelWindow::RegionFromDistrict(_district) == GW::Map::GetRegion()
-               && TravelWindow::LanguageFromDistrict(_district) == GW::Map::GetLanguage()
-               && (!_district_number || _district_number == static_cast<uint32_t>(GW::Map::GetDistrict()));
+            && GW::Map::GetMapID() == outpost_id
+            && TravelWindow::RegionFromDistrict(_district) == GW::Map::GetRegion()
+            && TravelWindow::LanguageFromDistrict(_district) == GW::Map::GetLanguage()
+            && (!_district_number || _district_number == static_cast<uint32_t>(GW::Map::GetDistrict()));
     }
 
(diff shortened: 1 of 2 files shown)


GWToolboxdll/Windows/TravelWindow.cpp
```diff
@@ -164,6 +164,53 @@ void TravelWindow::TravelButton(const char* text, const int x_idx, const GW::Con
...
+bool TravelWindow::PlayerHasAnyMissingOutposts(bool presearing) {
+    if (presearing) {
+        if (!GW::Map::GetIsMapUnlocked(GW::Constants::MapID::Ascalon_City_pre_searing))return true;
+        if (!GW::Map::GetIsMapUnlocked(GW::Constants::MapID::Ashford_Abbey_outpost))return true;
+        if (!GW::Map::GetIsMapUnlocked(GW::Constants::MapID::Foibles_Fair_outpost))return true;
+        if (!GW::Map::GetIsMapUnlocked(GW::Constants::MapID::Fort_Ranik_pre_Searing_outpost))return true;
+        if (!GW::Map::GetIsMapUnlocked(GW::Constants::MapID::The_Barradin_Estate_outpost))return true;
+    }
+    else {
+        for (int i = 0; i < (int)outpost_names.size(); ++i) {
```
franneck94, line 176: we want to use c++ cast only please

GWToolboxdll/Windows/TravelWindow.h
```diff
@@ -64,6 +64,8 @@ class TravelWindow : public ToolboxWindow {
     static GW::Constants::MapID IndexToOutpostID(int index);
     static bool ParseDistrict(const std::wstring& s, GW::Constants::District& district, uint32_t& number);
     static bool ParseOutpost(const std::wstring& s, GW::Constants::MapID& outpost, GW::Constants::District& district, const uint32_t& number);
+    bool PlayerHasAnyMissingOutposts(bool presearing);
```
franneck94, line 67: these params can be const

GWToolboxdll/Windows/TravelWindow.cpp
DubbleClick: preferably no cast and auto i = 0u;

GWToolboxdll/Windows/TravelWindow.h
DubbleClick: this is a declaration, they should be const in the definition only

franneck94: Is there a reason not having at also at the declaration?
I see no benefit in doing this

PR #947, opened 2023-10-10
feat: hide equipped and customized items

ChandlerFerry:
- Hides set equipped items from merchant
- Hides customized items from merchant

### DIFF
--- a/GWToolboxdll/Modules/InventoryManager.cpp
+++ b/GWToolboxdll/Modules/InventoryManager.cpp
@@ -52,6 +52,7 @@ namespace {
     };
 
     bool hide_unsellable_items = false;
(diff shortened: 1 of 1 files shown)


3vcloud:
Isn’t there a tab in the merchant window to hide your equipped items?

ChandlerFerry:
This functions as directly "equipped":
image

`item->equipped` is specifically meant to remove items that are equipped in a set that is not currently in use:
image

_Note the Zealous Gazing Scythe of Enchanting/Fiery Grim Scythe/Barbed Flatbow with icons identifying that they are equipped in a set._

DubbleClick:
Don't have time to check if this works without side effects, but assuming it does, it's a good addition.

3vcloud:
image

PR #948, opened 2023-10-12
feat: lock specific items

ChandlerFerry:
Right click -> lock/unlock **specific** items from:
- [x] Merchant
- [ ] Manual Salvage
- [x] Salvage All (Toolbox)
- [ ] Drop
- [ ] Trash

### DIFF
--- a/GWToolboxdll/Modules/InventoryManager.cpp
+++ b/GWToolboxdll/Modules/InventoryManager.cpp
@@ -54,6 +54,7 @@ namespace {
     bool hide_unsellable_items = false;
     bool hide_weapon_sets_and_customized_items = false;
     std::map<uint32_t, std::string> hide_from_merchant_items;
+    std::map<std::string, std::string> locked_items;
 
 
     bool GetIsProfessionUnlocked(GW::Constants::Profession prof)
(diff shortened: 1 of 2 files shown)


ChandlerFerry:
Still working on this; opened a draft PR to start receiving feedback.

1. How can I prevent Manual Salvage / Drop / Trash packets from being triggered when a locked item is the target?

2. What is a more optimal way to calculate a unique identifier for an item?
  - Currently merging the full mod_struct, is heavy

3vcloud:
No easy way to generate a true identifier for an item, it depends on use case e.g dye info, customised to etc. A full mod struct combined with a model id is the most sane way to do this for most scenarios, but if you don’t care about mods then name id would be another option

3vcloud:
Salvage, drop and trash actions (not packets) would need to be blocked further up the chain to avoid ui elements looking buggy, I’ll see if I can hook it in a good place at some point

ChandlerFerry:
I believe we will need to do full mod struct + model id to handle the use cases I see, I'll think about if name + model id works, but inscriptions & requirements aren't maintained in that case.


Noted on salvage/drop/trash actions, it's greatly appreciated.
Please let me know if there is anything I can do to assist.

ChandlerFerry:
For transparency, I won't be able to loop back to this until late November as I'm occupied preparing for WebSummit Lisbon.

DubbleClick:
no problem, feel free to finish it when you have the time

Issue #944, opened 2023-10-10
Minimap: symbols stop rendering on changing settings


adituv:
When any symbol setting is changed, including the Load Defaults button and the global Load Now button in settings, symbols will no longer render until Toolbox is restarted.  It seems that on the buffer being invalidated, the symbol renderer is never reinitialized due to the `if (initialized)` check.  I did a quick fix on my end that seems to work by creating an override for Invalidate that also sets initialized to true, but I am unsure if this is correct across the whole implementation, so thought it better to report in an issue.

PR #951, opened 2023-10-21
Fix SymbolRenderer reinitialisation

adituv:
#944 

VBuffer::Invalidate frees the dx buffer with an expectation that it will be reinitialized, but SymbolsRenderer did not reinitialise.  Now it will.


DubbleClick:
I fixed something like this about a month ago. I guess it's a thing at more places.

PR #957, opened 2023-10-26
Allow rendering all quest markers on minimap

### LINKED ISSUE #803: Show all quest log markers at once
Option to automatically cycle through quest log maybe on map change, grab quest indicator markers and plot on minimap to save having to cycle through all quests to find nearest one

adituv:
Adds an option to render all quest markers on the minimap rather than just the active quest

Closes #803

### DIFF
--- a/GWToolboxdll/Widgets/Minimap/Minimap.cpp
+++ b/GWToolboxdll/Widgets/Minimap/Minimap.cpp
@@ -351,6 +351,14 @@ void Minimap::OnUIMessage(GW::HookStatus*, const GW::UI::UIMessage msgid, void* 
                 SetWindowVisible(GW::UI::WindowID_Compass, true);
             }
             instance.is_observing = GW::Map::GetIsObserving();
+            // Cycle active quests to cache their markers
+            auto active_quest_id = GW::QuestMgr::GetActiveQuestId();
+            if (const GW::QuestLog* questLog = GW::QuestMgr::GetQuestLog(); questLog != nullptr) {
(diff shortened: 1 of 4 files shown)


3vcloud:
Id be very surprised if this worked, the game doesn’t give you all quest information for every quest until you switch to it

3vcloud:
I think I added something in the info window to get all quest info, you’ll need to factor that in

adituv:
It worked with my testing, but it was only with half a dozen quests in Plains of Jarin.  I'll have a further look.

3vcloud:
There’s a flag in the quest that tells you if it’s got all the data from the server or not, you can use that

3vcloud:
Do you even need to set as the active quest to get the marker info you need? Pretty sure that api call will sort it without messing with the users ui

adituv:
It didn't, unfortunately.  The active quest method was because I didn't want to reverse engineer to find a new call, but a method that doesn't affect the user's UI would definitely be preferable.

Testing:
* I used the button in Info window to request quest info, markers did not show up.  flags = 0x1
* I set quest active manually in the quest log, marker showed up.  flags still = 0x1
* Repeated for several other quests, and markers appeared one by one on the minimap

3vcloud:
Ok we’ll hold this until I take another look at the quest code

DubbleClick:
Into dev, please :)

adituv:
Ah, I swear that never used to be an option!  I've closed and reopened PRs to be against a different branch so many times by now...

PR #970, opened 2023-12-09
add lockpicks to pcons window

chypriote:
Lockpicks get regular use on almost every zone, think it makes sense to add them here

### DIFF
--- a/GWToolboxdll/Windows/PconsWindow.cpp
+++ b/GWToolboxdll/Windows/PconsWindow.cpp
@@ -107,6 +107,8 @@ PconsWindow::PconsWindow()
 
     pcons.push_back(new PconRefiller(L"Scroll of Resurrection", ItemID::ResScrolls, 5));
     pcons.back()->ini = "resscroll";
+    pcons.push_back(new PconRefiller(L"Lockpick", ItemID::Lockpick, 50));
+    pcons.back()->ini = "lockpick";
     pcons.push_back(new PconRefiller(L"Powerstone of Courage", ItemID::Powerstone, 5));
     pcons.back()->ini = "pstone";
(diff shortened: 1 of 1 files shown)


3vcloud:
Nope, you should instead make a hotkey to withdraw the item on map load for your use case. Whilst there are use cases for different types of summoning stones for certain speed clears, there’s no reason for lock picks to be added to the pcons window when there’s already a feature for withdrawing item by id

adituv:
I don't see why lockpicks wouldn't fit here.  Couldn't speedclearers also use the "withdraw item by id" feature for their summoning stones?  Can the "withdraw item by id" feature maintain a minimum number of lockpicks on a character?  Particularly for poorer players who don't necessarily have a stack of lockpicks per character, I think this could be a useful addition

DubbleClick:
This can be achieved with the /withdraw command too, so I agree that this isn't necessary. I'd even say that summoning stones have no business in the pcons window.

3vcloud:
Pcons are there primarily to be able to be turned on or off to automatically be used in an explorable- the autofill was added later as a nice to have, and has now also been separated out into the chat command for everything else. 

Everything currently in the pcons window is time sensitive to be renewed when it runs out in an explorable area, lock picks are not

chypriote:
Think it's a sensitive way of deciding what to add or not to that window. Should I update this to remove the few items that should not be here then ? I'm thinking of stones, scrolls and morale pcons

3vcloud:
Maybe this isn’t worth the effort to argue, I think maybe we should just add this in with a view to being able to add more items like this through the ui in future? If it’s not enabled by default it’s not adding cruft for existing users

chypriote:
I believe an UI similar to the one for hiding items from merchant would be the most fitting. Would you prefer closing that one and reopening a PR after I've implemented that ?

PR #986, opened 2024-03-17
Fix crashes from GetRemainingArgsWStr

adituv:
GetRemainingArgsWstr previously returned nullptr when there was no space within the input string, causing crashes if e.g. the chat command "/show" was used with no parameter given.

This adds an extra check so GetRemainingArgsWstr instead returns an empty string in that case.  Depending on intended behaviour, it may instead be preferable to add null checks to every call site of GetRemainingArgsWStr.


adituv:
Ok, renaming the branch closes the pull request, noted for future.  My bad.

PR #1040, opened 2024-04-26
Fix null pointer crash in quest marker renderer

adituv:
`active_quest` is null when the current mission's description is selected in the quest log.

### DIFF
--- a/GWToolboxdll/Widgets/Minimap/SymbolsRenderer.cpp
+++ b/GWToolboxdll/Widgets/Minimap/SymbolsRenderer.cpp
@@ -182,7 +182,8 @@ void SymbolsRenderer::Render(IDirect3DDevice9* device)
     const auto draw_quest_marker = [&](const GW::Quest& quest)
     {
         const auto active_quest = GW::QuestMgr::GetActiveQuest();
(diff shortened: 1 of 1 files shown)


DubbleClick:
good find, but now the question is whether you can have quests in the log, but not an active one?

if that's the case, we should still draw all quests, regardless of whether there is an active one

3vcloud:
Could get messy, the ui is reactive but ours is immediate - we may have to hook the compass function or frame to find out the actual value and then it’s not worth the feature

adituv:
> good find, but now the question is whether you can have quests in the log, but not an active one?

Good call, and the answer is yes. I managed to get this to happen by making a new character in Pre-Searing, taking Message from a Friend and Warrior Test, then selecting Message from a Friend in the quest log and completing that quest.  Warrior Test was left in the log but not active.

DubbleClick:
yep agreed, just don't do the check with the active quest if it's null

adituv:
There's also a null pointer crash in the filter lambda, where quest is somehow null despite being a reference.   Still working on this.

adituv:
Done.  Turns out the active quest id for the current mission is -1 so the comparison to `QuestID::None`  was insufficient.

PR #1053, opened 2024-05-02
Salvage information module

### LINKED ISSUE #1052: Add salvaging info to items
When hovering over an item it would indicate which materials it salvages into. This could be limited to common materials to avoid clogging up the UI and help with con set focused materials.

Color coding could help distinguish things at a glance, similar to how the price of runes are color coded with rarities.

AlexMacocian:
Closes #1052


AlexMacocian:
Stilll needs to figure the encrypted strings reverse lookup

AlexMacocian:
Works with fully localized message

PR #1072, opened 2024-05-10
Refactor item description hook into its own module

adituv:
Adds a mandatory module that provides a hook to modify the generation of item names and descriptions.
Refactors the existing disparate hooks to use the new functionality.

### DIFF
--- a/GWToolboxdll/Modules/ItemDescriptionHandler.cpp
+++ b/GWToolboxdll/Modules/ItemDescriptionHandler.cpp
@@ -97,3 +97,20 @@ void ItemDescriptionHandler::OnGetItemDescription(const uint32_t item_id, const 
 
     GW::Hook::LeaveHook();
 }
+
+// Unregisters all instance of the provided callback
+void ItemDescriptionHandler::UnregisterDescriptionCallback(ItemDescriptionCallback callback)
+{
+    erase_if(this->callbacks, [callback](const auto& cb) {
+        return cb.callback == callback;
+    });
+}
+
+// Unregisters all instances of the provided callback at the provided altitude.  Leaves all instances at other
+// altitudes registered.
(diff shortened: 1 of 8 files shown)


GWToolboxdll/Modules/ItemDescriptionHandler.cpp
```diff
@@ -78,15 +78,15 @@ void ItemDescriptionHandler::OnGetItemDescription(const uint32_t item_id, const
         }
 
         if (out_name != nullptr) {
-            if (args.block_name) {
```
adituv: As a side-note, I used `.c_str()` because I wasn't sure whether dereferencing an empty wstring directly would be undefined behaviour.

> If pos is equal to the [string length](https://cplusplus.com/basic_string::length), the const-version never throws exceptions (no-throw guarantee).
Otherwise, it causes undefined behavior.

GWToolboxdll/Modules/ItemDescriptionHandler.cpp
```diff
@@ -78,15 +78,15 @@ void ItemDescriptionHandler::OnGetItemDescription(const uint32_t item_id, const
...
+            if (args.block_name || modified_name.c_str()[0] == L'\0') {
```
adituv, line 81: As a side-note, I used `.c_str()` because I wasn't sure whether `modified_name[0]` would be UB for an empty wstring.

> If pos is equal to the [string length](https://cplusplus.com/basic_string::length), the const-version never throws exceptions (no-throw guarantee).
Otherwise, it causes undefined behavior.

\- https://cplusplus.com/reference/string/basic_string/operator[]/

DubbleClick: You can use string.empty()? It's not functionally equivalent since a string could technically begin with a null terminator, but we'd have to really mess up in generating the string for that to be an issue.

PR #1084, opened 2024-05-17
Salvage Info Module Fixes

AlexMacocian:
Add missing Scale material
Add timeout logic for failed fetches
Added failure message on request failures
Add a more lax string comparison, to match on plurals, case-insensitive and special handling for Ruby/Rubies
Adjusted response caching to cache in the case of non-transient failures (codes from 300 to 499)
Lowered cache duration for salvage info to 1 day. This way the users can get fresh info once I add new wiki redirects for problematic items

### DIFF
--- a/GWToolboxdll/Modules/Resources.cpp
+++ b/GWToolboxdll/Modules/Resources.cpp
@@ -617,7 +617,9 @@ void Resources::Download(const std::string& url, AsyncLoadMbCallback callback, v
         }
         std::string response;
         bool ok = Download(url, response);
-        save_to_cache(cache_path, response);
+        if (ok) {
+            save_to_cache(cache_path, response);
+        }
         EnqueueMainTask([callback, ok, response, context] {
             callback(ok, response, context);
         });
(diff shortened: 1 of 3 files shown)


GWToolboxdll/Modules/Resources.cpp
3vcloud, line 1: Need to expand the callback arguments for this a bit more; for example, an error 404 is actually pretty valid for the cache period - you would want to cache that response to reduce network requests anyway

AlexMacocian: True. I'll look into it

AlexMacocian: Okay so I'll add an overload that gives you the status code of the request and for caching, I'll make it cache for all responses between 200 to 499.
200s are successes
300s are redirections
400s are client errors - these ones specifically can be cached because a retry won't result in a different outcome

AlexMacocian: Done

PR #1088, opened 2024-05-19
Implement active quest objectives widget

adituv:
Creates ActiveQuestWidget, which displays the name and objectives of the current active quest.

This ended up way more complicated than I wanted, largely due to handling mission objectives (and greying out completed objectives).  But also in general because I needed custom callbacks on string decoding so couldn't use EncString for most of it.


adituv:
Other ideas: potentially needs some way to handle long lists of objectives.  A scrollbar (I guess convert it to a window for this?)?  Swap the objective order, as uncompleted objectives are more likely to be lower?  A setting to hide completed objectives?

PR #1110, opened 2024-06-14
6.21

DubbleClick:
compass fix


BearsMan:
This version isn't pushed on the website and more. I thought I would add the comment for that here.

DubbleClick:
Well aware. Not happening yet.

PR #1147, opened 2024-07-04
Replace magic numbers in EnemyWindow

franneck94:
Just scrolled though the code.

But do we really want to filter these out from the list?

image

Or rather add an setting for ModelIDs to filter out?


3vcloud:
It’s probably a hangover from the original logic that I missed - this should be removed anyway to simplify the module, and if it’s an issue later we can revisit adding yet another set of options

PR #1149, opened 2024-07-05
Nic item collection date

### LINKED ISSUE #1148: What date nicholas wants a trophy
It might be more useful tooltip to say what date nicholas wants a trophy, instead of how many he wants.

So instead of:
"Nicholas wants {X} of these!"
It could be:
"Nicholas wants these next on {Date}!"
or
"Nicholas wants these in {X} weeks!"
or
"Nicholas wants {X} of these in {Y} weeks!"

AlexMacocian:
Closes #1148


3vcloud:
Look into using the encoded string that’s used in the guild window to automatically provide the time string - you’ll need to take a look at the enc string log in the info window when opening the guild window

AlexMacocian:
Sure I'll take a look at it later in the evening

PR #1150, opened 2024-07-05
added nic item in n weeks to desc string

franneck94:


franneck94:
Since alex also pushed it, this can be closes i guess

3vcloud:
Thanks for jumping on this one anyway :)

Issue #1183, opened 2024-07-24
Draw paths to objectives


AlexMacocian:
Use the pathfinding implementation in toolbox to obtain paths to quest objectives, then draw those paths over the toolbox minimap

AlexMacocian:
I know that the pathfinding module is not completely functional yet, but leaving this as a reminder. I'll try to tackle it next week once I get home from holidays

AlexMacocian:
Oh I see the logic already exists in the QuestModule, but currently the module is not hooked up anywhere

3vcloud:
More logic needs to be written around how and when to recalculate the path based on the players position between points and the end point. You can turn it on with a few lines of code, and I’m sure you’ll come across the same issue I did

AlexMacocian:
Idk why the custom renderer doesn't draw the lines for the waypoints, but I see the issue with the quest markers

AlexMacocian:
So I think the reason the logic keeps spamming marker updates is because, first you call calculate, which sets the current waypoint to 0, so the logic updates the marker to the waypoint 0 (that is the player position).
Then, almost immediately, Update() calculates the correct waypoint, detects that the waypoint is different than waypoint 0 and Updates the UI

3vcloud:
Taken discussion onto discord

PR #1185, opened 2024-07-24
Quest objective pathfinding fixes

AlexMacocian:
Deduplicate UI updates on path calculation
Reduce sensibility to waypoint updates to reduce UI update spam
Addresses #1183

### DIFF
--- a/GWToolboxdll/Modules/QuestModule.cpp
+++ b/GWToolboxdll/Modules/QuestModule.cpp
@@ -29,6 +29,7 @@ namespace {
         }
         std::vector<GW::GamePos> waypoints;
         std::vector<CustomRenderer::CustomLine*> minimap_lines;
+        GW::GamePos previous_closest_waypoint;
         GW::GamePos original_quest_marker;
         GW::GamePos calculated_from;
         GW::GamePos calculated_to;
@@ -89,7 +90,7 @@ namespace {
                 Recalculate(from);
                 return false;
             }
-            constexpr float recalculate_when_moved_further_than = 1500.f * 1500.f;
+            constexpr float recalculate_when_moved_further_than = 100.f * 100.f;
             if (GetSquareDistance(from, calculated_from) > recalculate_when_moved_further_than) {
                 Recalculate(from);
                 return false;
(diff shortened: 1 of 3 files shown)


GWToolboxdll/Modules/QuestModule.cpp
```diff
@@ -119,8 +120,15 @@ namespace {
             if (waypoints.empty())
                 return;
             DrawMinimapLines();
-            if(IsActive())
+            const auto& current_waypoint_pos = waypoints[current_waypoint];
+            const auto waypoint_distance = GetSquareDistance(current_waypoint_pos, previous_closest_waypoint);
+            constexpr float update_when_waypoint_changed_more_than = 300.f * 300.f;
```
AlexMacocian, line 125: This part tries to avoid updating the quest marker as long as the current marker is "sufficiently accurate"

GWToolboxdll/Modules/QuestModule.cpp
```diff
@@ -217,6 +225,24 @@ namespace {
             // Waypoint array is in descending distance, flip it
             std::reverse(cqp->waypoints.begin(), cqp->waypoints.end());
         }
+
+        const auto waypoint_len = waypoints.size();
```
AlexMacocian, line 229: This part was missing from the original implementation. Without this, the UI would be updated to point to the player and then the immediate Update() call would recalculate the waypoints properly, causing the UI to update again to point to the correct waypoint.

GWToolboxdll/Modules/QuestModule.cpp
```diff
@@ -89,7 +90,7 @@ namespace {
                 Recalculate(from);
                 return false;
             }
-            constexpr float recalculate_when_moved_further_than = 1500.f * 1500.f;
+            constexpr float recalculate_when_moved_further_than = 100.f * 100.f;
```
AlexMacocian, line 93: Lowered the recalculation period to have the waypoints a bit more smooth. This can be reverted to improve performance

3vcloud: I haven't messed with this in a while, whats the delay like on debug?

AlexMacocian: It's like ~100ms so not really that noticeable per path, but depends heavily on the map

PR #1197, opened 2024-07-30
PathfindingWindow improvements

AlexMacocian:


GWToolboxdll/Modules/QuestModule.cpp
```diff
@@ -121,9 +124,23 @@ namespace {
...
+                return false;
+            }
             if (calculating) {
                 return false;
             }
+            if (!updatingQuestMarker && original_quest_marker.y == INFINITY) {
+                updatingQuestMarker = true;
+                GW::GameThread::Enqueue([this]{
+                    const auto quest = GW::QuestMgr::GetQuest(quest_id);
+                    ASSERT(quest);
```
AlexMacocian, line 138: I'm afraid about this assert a bit. Can it trigger if a quest is finished while on the map?

PR #1199, opened 2024-07-31
Add stop_token to cancel pathfinding

AlexMacocian:


AlexMacocian:
Don't merge it yet, found one crash in Release and I'm trying to replicate it now

GWToolboxdll/Modules/QuestModule.cpp
```diff
@@ -233,11 +245,13 @@ namespace {
     void OnQuestPathRecalculated(const std::vector<GW::GamePos>& waypoints, void* args)
     {
         const auto cqp = GetCalculatedQuestPath(*(GW::Constants::QuestID*)&args, false);
-        if (!(cqp && cqp->calculating)) 
+        if (!cqp) 
             return;
         // TODO: @3vcloud idc to look at it atm but this crashes me when changing zones if a path had already been calculated
-        ASSERT(cqp->calculating);
+        ASSERT(cqp->stop_source.has_value());
```
AlexMacocian, line 251: I suspect this is causing the crashes, but I wasn't able to reproduce it yet

AlexMacocian:
Tried zoning in and out of towns for like 10 minutes straight, with RelWithDebInfo, wasn't able to trigger any other crash.

Only crash I've had was once with an error that seems unrelated to the changes in this PR.
image

I'd say it can be merged

GWToolboxdll/Windows/Pathfinding/PathfindingWindow.cpp
```diff
@@ -257,6 +257,10 @@ bool PathfindingWindow::CalculatePath(const GW::GamePos& from, const GW::GamePos
             goto trigger_callback;
         }
 
+        if (stop_token.stop_requested()) {
```
DubbleClick, line 260: am I missing something or does this only stop the calculation before it actually starts in Resources::EnqueueMainTask?

AlexMacocian: Yeah, we could stop it altogether but I didn't want to touch the inner logic for now. The idea was just to not do the callback which would call an already deleted object.

I'll adjust the pathfinding later to actually cancel the pathfinding algorithm

DubbleClick:
Sorry for forcing the rebase, seeing those memory allocations drove me mad 🤪

AlexMacocian:
Ah fk I thought it was merged so I deleted the branch xD...

GWToolboxdll/Windows/Pathfinding/PathfindingWindow.cpp
AlexMacocian: Propagated the token all the way to Pathing

PR #1226, opened 2024-09-02
Updated get_next_bag_slot() to account for a PR in GWCA

zhro:
Updated get_next_bag_slot() to account for a PR in GWCA which fixes a bug in the GW::Constants::Bag prefix increment operator overload. This PR can be pulled prior to the GWCA PR, as the behavior is unchanged. The prefix increment simply had to be moved to a non-const object, as the increment operator now returns a reference (as expected) instead of a copy. The prefix operator is expected to modify the variable is it being used upon, which is invalid for a const object.

### DIFF
--- a/GWToolboxdll/Modules/InventoryManager.cpp
+++ b/GWToolboxdll/Modules/InventoryManager.cpp
@@ -1502,7 +1502,8 @@ bool get_next_bag_slot(const InventoryManager::Item* item, GW::Constants::Bag* b
     if (slot >= bag->items.size()) {
         bag_id = GW::Constants::Bag::Max;
         slot = 0;
-        for (auto it_bag_id = ++bag->bag_id(); it_bag_id < GW::Constants::Bag::Max; it_bag_id++) {
+        auto bag_start = bag->bag_id();
+        for (auto it_bag_id = ++bag_start; it_bag_id < GW::Constants::Bag::Max; it_bag_id++) {
             const auto it_bag = GW::Items::GetBag(it_bag_id);
             if (it_bag) {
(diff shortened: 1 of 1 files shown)


DubbleClick:
This PR does nothing except declare a variable. Will merge the GWCA PR though

DubbleClick:
Oh was the GWCA PR already merged?

zhro:
Not yet. Once GWCA merges the corrected operator overload, this piece of code will fail to compile without this change.

Here is the referenced PR:

https://github.com/gwdevhub/GWCA/pull/11

DubbleClick:
Ok perfect, looks like I wasn't subscribed to our  forked GWCA repo.

Should add a operator++(Bag&&) that returns a copy, then this spaghetti here is not required.

zhro:
It's not spaghetti. The way that it's written currently makes no sense. Take this example:

```
auto a = ++b;
```

One would assume that this code increments b and returns a reference to the same object. But that's not what the problem was here. The issue was this:

```
const b;

auto a = ++b;
```

You can't increment a const and adding an overload that allows this to work is going to create one of two problems: either you don't increment `b` (which was the source of the original problem) or you increment b, which violates the contract of `b` being const.

You have to make a copy to a non-const object explicitly before the prefix increment operator can be called, because it is modifying the object.

If you want a copy without assignment then the addition operator should be overloaded instead. The new (valid) code would look like this:


```
const b;

auto a = b + 1;
```

DubbleClick:
Yeh then that's fine too. I'll do it when I merge in the gwca change later

PR #1235, opened 2024-09-27
Do not block /tb commands from reaching TB

JaborGW:
Currently if any UI plugin is injected and the user sends a `/tb` chat command, this command is blocked by the plugin and does not reach toolbox.
To fix this, chat commands are now only blocked if they start with `/tb <pluginname>`


DubbleClick:
that had to be fixed on the plugin side, not in the project itself. example plugin already doesn't block /tb

PR #1236, opened 2024-09-27
Do not block /tb commands from reaching TB

JaborGW:
Currently if any UI plugin is injected and the user sends a `/tb` chat command (for example `/tb settings`) this command is blocked by the plugin and does not reach toolbox. That is because the default `blocked` state of the hook is `true`, meaning all the existing early-return cases of `CmdTB` block the chat command.

To fix this the behaviour is inverted: By default, plugins let the chat commands through. The only chat commands that are blocked are those that start with `/tb <pluginname>`.

### DIFF
--- a/plugins/Base/ToolboxUIPlugin.cpp
+++ b/plugins/Base/ToolboxUIPlugin.cpp
@@ -11,23 +11,21 @@
 namespace {
     void CmdTB(GW::HookStatus* status, const wchar_t*, const int argc, const LPWSTR* argv)
     {
+        status->blocked = false;
         const auto instance = static_cast<ToolboxUIPlugin*>(ToolboxPluginInstance());
         if (!instance) {
-            status->blocked = false;
             return;
         }
         if (argc < 3) {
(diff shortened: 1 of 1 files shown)


plugins/Base/ToolboxUIPlugin.cpp
```diff
@@ -11,23 +11,21 @@
...
     {
+        status->blocked = false;
         const auto instance = static_cast<ToolboxUIPlugin*>(ToolboxPluginInstance());
         if (!instance) {
-            status->blocked = false;
             return;
         }
         if (argc < 3) {
-            status->blocked = false;
+            return;
```
JaborGW, line 20: This extra early return is unrelated, but seems like it's missing to me: I havent tested it, but considering we access `argv[2]` later down the line for the chat command `/tb <pluginname>`, that seems like it would cause crashes.

JaborGW:
Sorry, looks like this is already fixed by on `dev`, but not on `master`; I didnt realize when fixing the merge conflicts.

PR #1238, opened 2024-09-30
remove settings window, make quiet default, do not update toolbox if …

DubbleClick:
…it's installed

### DIFF
--- a/GWToolbox/Install.cpp
+++ b/GWToolbox/Install.cpp
@@ -112,7 +112,7 @@ bool DeleteInstallationDirectory()
     return true;
 }
 
-bool Install(const bool quiet)
+bool Install()
 {
     if (IsInstalled()) {
         return true;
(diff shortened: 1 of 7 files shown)


GWToolbox/Install.cpp
```diff
@@ -133,36 +133,17 @@ bool Install(const bool quiet)
...
-            L"Do you want to delete *all* possible files from installation folder? (Default: no)\n",
-            L"Uninstallation",
-            MB_YESNO);
-
-        if (iRet != IDYES) {
-            DeleteAllFiles = false;
-        }
-    }
+    // Delete all files
+    DeleteInstallationDirectory();
```
DubbleClick, line 144: not entirely sure about this, but if the user clicked no, it essentially didn't do anything

GWToolbox/main.cpp
```diff
@@ -176,19 +167,12 @@ int WinMain(HINSTANCE hInstance, HINSTANCE hPrevInstance, LPSTR lpCmdLine, int n
 
         if (iRet == IDYES) {
             // @Cleanup: Check return value
-            if (!Install(settings.quiet)) {
+            if (!Install()) {
```
DubbleClick, line 166: I think it's fair to always show the box here

PR #1245, opened 2024-10-18
/pc results increased to 12 and party search text save between sessions

raben1:


3vcloud:
So I was thinking about this one - should the /pc command instead open the trade window with the search term instead? You’re expecting to engage with the content at that point so interrupting gameplay by showing a window wouldn't be a concern, and this way it doesn’t clog up the chat log

raben1:
I find the /pc command really useful for a quick search, like in the middle of a run or waiting to fill the team, then you can open the trade window if you need a deeper search results. You can yet interact with the /pc results since the users are targetable to pm.

DubbleClick:
wtf is the /pc command?

BearsMan:
/pc if I am right, it means price check.

PR #1246, opened 2024-10-18
Move to trade on double click or alt+click or disabled

raben1:
Fixed ctrl+click not moving id kits nor salvage kits (they were poping context menu).

*Any kind of feedback would be really appreciated on this or other PR's, please.*


GWToolboxdll/Modules/GameSettings.cpp
```diff
@@ -1986,6 +1994,12 @@ void GameSettings::DrawInventorySettings()
     ImGui::TextDisabled(logic);
     ImGui::Unindent();
     ImGui::Unindent();
+
+    ImGui::Checkbox("Move items to trade on Double Click", &move_to_trade_on_double_click);
```
3vcloud, line 1998: There’s a radio group option for this type of thing - take a look at some of the features that allow you to do stuff in outpost/explorable

Issue #1326, opened 2025-01-26
More and flexible Transmo Options


Lennystair:
Hey, I love the Transmo / Transmoparty feature! Unfortunately it's very limited to the local NPC's in the instance or to a small selection of characters: "Kanaxai, Khunavang, Bone Dragon, Zhu..." I would love to make little lore in-game videos like this (about the Battle of Ascalon , the Guild Wars with Orr) as King Adelbern for example. Or standard NPC transmogs like "Ascalon Soldier" or"Dwarven Soldier". Is there a reason why the characters are limited? Is there a way to add more here? I've seen it all in Guild Wars, I'd like to do more roleplaying videos here,  Thank you! :)

DubbleClick:
It's limited because we have to manually add npcs. If there are some you'd like, you may add them here: https://github.com/gwdevhub/GWToolboxpp/blob/f16bfe8fae1c65d0fc98dc9ce82dc9156435fa82/GWToolboxdll/Modules/ChatCommands.cpp#L1202

Lennystair:
@DubbleClick  that doesn't sound too difficult at all... but unfortunately i'm not that into the topic to be honest. Maybe someone will take care of implementing some characters in the future. Thanks anyways :)

Irisier:
would it be possible to make the fields for NPC ID, NPC ModelfileID, NPC Modelfile, and NPC Flags variable for /transmo && /transmoparty via a corresponding input, like def_scale? then you could use any model if you only know the corresponding IDs.

3vcloud:
@claude make the transmo list dynamic so a user can manually add, amend or remove entries. Add a reset button to revert the entries to the current set as a default. Also add a button "Add current target" that will automatically add the current target as an entry, decoding the target's encoded name via EncString as the entry name.

Also add "/transmo add" that would do the same thing as clicking the "add current target" button would do.

3vcloud:
@claude looks like transmo settings aren't being saved/loaded properly, can you take a look?

PR #1345, opened 2025-02-18
feat: transmo subcommand to load supplied npc model id

Aelto:
As mentioned in #1326, the current list of NPCs to transmo into is quite limited, and manually adding NPC models to the hardcoded list can be a complicated task for those without cpp knowledge. This PR aims to add a subcommand to the transmo command: `/transmo model`. This subcommand accepts every parameter needed by the `PendingTransmo` struct, allowing anyone to transmo into the NPC model of their choice without having to edit the source code.

The command asks for specific IDs to function correctly which is pretty verbose to type, so it is mostly here to be used through chat command aliases or hotkeys.

Examples:
```sh
# transmo as Gehraz
/transmo model 4581 204020 245127 540

# transmo as Herta
/transmo model 4537 16258 245122 540

# transmo as Anzhaer
/transmo model 7985 16271 373574 520
```

Also as the command requires a lot of parameters it felt adequate to add a lot of error handling and logs, but if it's too much feel free to tell me!

### DIFF
--- a/GWToolboxdll/Modules/ChatCommands.cpp
+++ b/GWToolboxdll/Modules/ChatCommands.cpp
@@ -2719,6 +2719,42 @@ void CHAT_CMD_FUNC(ChatCommands::CmdTransmo)
         if (wcsncmp(argv[1], L"reset", 5) == 0) {
             transmo.npc_id = std::numeric_limits<int>::max();
         }
+        else if (wcsncmp(argv[1], L"model", 5) == 0) {
+            bool invalid_data = argc < 6;
+            int npc_id = transmo.npc_id;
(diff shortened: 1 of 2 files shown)


Irisier:
Hey, could we please add NPC Scale to the command? There are some models in the game that are nicer to use scaled.

3vcloud:
See chat command help

Irisier:
ok, thanks, I didn't know that /transmo can be applied to yourself with size even without a selected target

Lennystair:
The examples like Gehraz works great, but many others are displayed as white cubes. Or am I doing something wrong with the hexadecimal to decimal conversion? Can you give me the code f.e. for King Adelbern?

PR #1356, opened 2025-02-25
Keyboard shortcuts for salvage all item list modal

Aelto:
Hi again, I hope you don't mind these small and very focused PRs as I try to make things easy for you to review.

Related to #1355, adds keyboard shortcuts to the item selection modal that appears while trying to _Salvage All_.

- The `Ok` button can now be triggered by pressing either `Space` or `Enter`
- `Escape` can perform two actions:
  - if there is one or more items selected then pressing the key will deselect them all as if the `Select all` checkbox was clicked
  - if there is no item selected then pressing the key will close the modal as if the `Cancel` button was clicked

These shortcuts help for the following use cases:
- quickly salvaging everything without filtering ~> Press `Space` as soon as the modal appears
- salvaging only a few select items ~> Press `Escape` to unselect all, select the right items then press `Space` to confirm
- closing the modal to cancel ~> Press `Escape` once to unselect all, press a second time to simulate a click on `Cancel`

**Note:** _(solved, see commits below)_
I didn't find an easy way to suppress the keypress events for the Guild Wars client, so pressing `Space` to confirm the salvage will also cause the player to run towards or attack its target, or pressing `Escape` to cancel will also close the inventory panel. In my opinion it's an acceptable flaw, but if there is a simple way to do it for this piece of code then i'll gladly implement it.

### DIFF
--- a/GWToolboxdll/Modules/InventoryManager.cpp
+++ b/GWToolboxdll/Modules/InventoryManager.cpp
@@ -2018,9 +2018,17 @@ void InventoryManager::Draw(IDirect3DDevice9*)
                 }
                 ImGui::SameLine();
             }
-            if (ImGui::Button("Cancel", btn_width) || ImGui::IsKeyDown(ImGuiKey_Escape)) {
+            // Pressing [Escape] when no item is selected results in the window being closed.
+            // Or pressing [Escape] when some items are selected results in everything being unselected.
+            if (ImGui::Button("Cancel", btn_width) || ImGui::IsKeyPressed(ImGuiKey_Escape) && !has_items_to_salvage) {
                 CancelSalvage();
                 ImGui::CloseCurrentPopup();
+            }
+            else if (check_all_items && ImGui::IsKeyPressed(ImGuiKey_Escape)) {
+                check_all_items = false;
+                for (size_t i = 0; i < potential_salvage_all_items.size(); i++) {
+                    potential_salvage_all_items[i]->proceed = false;
+                }
             }
         }
         ImGui::EndPopup();
--- a/GWToolboxdll/Modules/InventoryManager.cpp
+++ b/GWToolboxdll/Modules/InventoryManager.cpp
@@ -1941,7 +1941,7 @@ void InventoryManager::Draw(IDirect3DDevice9*)
         else if (is_salvaging_all) {
             // Salvage in progress
             ImGui::Text("Salvaging items...");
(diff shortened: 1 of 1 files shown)


DubbleClick:
thank you very much, if you want to block the keypress being forwarded gw you have to return trie in wndproc. Look at the hotkey option to get a better understanding

Aelto:
Thanks for the hint, I've found the solution but I would like your opinion on how to approach it.

It looks like ImGui has a concept of `io.WantCaptureKeyboard` which is true when a modal is being displayed or when set via `SetNextFrameWantCaptureKeyboard(true)`. In the current state it seems that Toolbox doesn't read that flag in the [global WndProc](https://github.com/gwdevhub/GWToolboxpp/blob/master/GWToolboxdll/GWToolbox.cpp#L548-L550)
```cpp
if (io.WantTextInput) {
    return true; // if imgui wants them, send just to imgui (above)
}
```
Changing the `if` to the following enables keyboard capture globally for all modals:
```cpp
if (io.WantTextInput || io.WantCaptureKeyboard) {
    return true; // if imgui wants them, send just to imgui (above)
}
```

But i'm not sure if the missing check wasn't intentional in the first place, so perhaps doing it on a per-module basis with an explicit check at the end of WndProc would be better? [here in the case of InventoryManager](https://github.com/gwdevhub/GWToolboxpp/blob/master/GWToolboxdll/Modules/InventoryManager.cpp#L1013)
```cpp
// simplified for the snippet, we probably want mouse inputs to pass-through 
return ImGui::GetIO().WantCaptureKeyboard;
```

So in your opinion where should I add the check?
- in the global WndProc
- locally in the InventoryManager WndProc

Thanks in advance!

DubbleClick:
It's not really a missing check. Unless I'm mistaken, WantCaptureKeyboard is always true when WantTextInput is true. The latter exists as a distinct flag to let users react to typing scenarios and pop-in onscreen keyboards or such.

Aelto:
> Edit: oh wait, we're checking for the wrong flag I guess. It might have a reason then, I can't tell atm.

Indeed, the ImGui FAQ seems to recommend using `WantCaptureKeyboard` to block inputs so this must be on purpose. I've decided to handle that case in the inventory manager WndProc where it's guaranteed not to cause unintended input blocking.

DubbleClick:
thanks I'll take it out of inventorymgr and use it in the global wndproc instead

Issue #1366, opened 2025-03-10
Add navigation support for Proph desert portals


shiburito:
Not sure how possible it'll be to do but when working on my solo character I noticed the navigation doesn't work through desert portals in proph. Would be cool if it could lead you to them when the quest is on the other side

DubbleClick:
I'm not sure if thise is realistic for now. It won't be a priority, sorry.

shiburito:
Yep I figured low hanging but just wanted to document it before I forgot about it! Appreciate it!

544449:
someone just needs to map portal pairs like in the example:
[https://github.com/gwdevhub/GWToolboxpp/blob/57dcce0174160ba06ce291135a045004a49431a2/GWToolboxdll/Windows/Pathfinding/MapSpecificData.h#L36-L37](url)

PR #1369, opened 2025-03-11
feat(mapping) add prophets path portals

shiburito:
Implements part of #1366 - Isn't all areas but is a start


3vcloud:
> Please use z layer of portal plane and not zero.

image

Any idea on how we can get the zplane from the prop struct?

3vcloud:
Anyway, setting zplane to 0 should be fine but I don't think the pathing calc is working for portals at the moment

PR #1371, opened 2025-03-16
Pet Module - Adds feature to keep open the pet command window

shiburito:
Tested going in a mission
Tested going in and out of an outpost
Tested going in and out of two explorable areas

I didn't see a documented window ID for the pet command window so did it through a keypress, if there is a better way feel free to change it or lmk and I'm happy to.

### DIFF
--- a/GWToolboxdll/GWToolbox.cpp
+++ b/GWToolboxdll/GWToolbox.cpp
@@ -47,6 +47,8 @@
 
 #include <DelayImp.h>
 
+#include "Modules/PetModule.h"
+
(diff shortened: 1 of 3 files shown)


3vcloud:
Did you try using a hotkey to open the pet command window on map load?

shiburito:
I'm not sure how I have never seen that you could have hotkeys run on entering an outpost/explorable despite having used them for so long LOL. So yeah this can be done that way instead and this can be disregarded unless you have any other pet related features that you can think of that makes sense in here

PR #1375, opened 2025-03-24
Removed a check for PreSearing Region

KhiosDev:
When in Pre-Searing and using the "Find nearest outpost" function, it failed silently


3vcloud:
https://github.com/gwdevhub/GWToolboxpp/commit/5ac25092511e060b8c30fdddd47540bd219eb4b0

PR #1407, opened 2025-04-26
Export Resources::GetSkillImage for External Plugin Use

kd209:
This PR addresses an accessibility issue encountered while developing an external observer plugin that requires skill icon textures.

The `Resources::GetSkillImage` function, which provides access to skill textures, was previously internal to GWToolboxdll.dll. External plugins loaded as separate DLLs could not directly call this function, preventing them from displaying skill icons using the core Toolbox resources. 

I'm currently looking to implements Widgets and use these features to my Observe Plugin.

To resolve this, this PR introduces the following line to Resources.cpp:
```c++
#pragma comment(linker, "/EXPORT:?GetSkillImage@Resources@@SAPAPAUIDirect3DTexture9@@W4SkillID@Constants@GW@@@Z")
```

Its make possible to use this previous intern function on my DLL.

I tested and seems to not affected others parts of TB, and works on my side !

I hope this can be merged or a similar solution, so we can use textures on external plugins.

Thanks

image

### DIFF
--- a/GWToolboxdll/Modules/Resources.cpp
+++ b/GWToolboxdll/Modules/Resources.cpp
@@ -1052,6 +1052,8 @@ IDirect3DTexture9** Resources::GetGuildWarsWikiImage(const char* filename, size_
     return texture;
 }
 
+#pragma comment(linker, "/EXPORT:?GetSkillImage@Resources@@SAPAPAUIDirect3DTexture9@@W4SkillID@Constants@GW@@@Z")
+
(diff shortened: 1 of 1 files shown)


3vcloud:
Yep this is a good idea, I'll see what other functions would be good to export for third party access

3vcloud:
6ede592c3135e1a99faf35a6af813a2f1ca6cc4d

henderkes:
Since when are we exporting toolbox functionality for plugins? Shouldn't we move the resource stuff into gwca instead?

3vcloud:
We should yes

PR #1442, opened 2025-06-09
Resources export(GetMapName & DecodeStringId)

smn1337:
Hey, 
I'm asking to export 2 more functions, mainly to being able to decode Map Name on the game.

This is related to https://github.com/gwdevhub/GWToolboxpp/pull/1407


3vcloud:
You should instead be writing your own GWCA level calls to get this info - getting a resource image was a bit more complicated which is why it was exported

PR #1473, opened 2025-07-17
fix more evil msvc evil magic

cruel-efficiency:
I couldn't get it to build for the life of me until I changed these. Noticed there were similar commits in the repo.


henderkes:
already fixed in dev branch, but you're right, master should probably build too

PR #1505, opened 2025-11-29
Add comma-separated item IDs support to Use Item hotkey, E.G. for using a random miniature or tonic

Le-Vagabond-gh:
- Replace single item_id with vector<uint32_t> item_ids for multiple IDs
- Add comma-separated input field in UI with validation
- Implement random selection from item list on execution
- Maintain backwards compatibility with existing single item hotkeys
- Update save/load format to use ItemIDs field while preserving legacy support
- Remove restrictive ImGui flags that prevented comma input
- Keep using existing 'ItemID' field name for backwards compatibility
- Fix loading of comma-separated values from ini file
- Maintain single value backwards compatibility with 'item_id' fallback

Adjust Use Item hotkey UI layout and labels

- Increase item width from 140 to 250 pixels for better text visibility
- Simplify label from "Item Model IDs (comma-separated)" to "Item Model IDs"
- Update help text to clarify comma-separated format and random selection
- Change "Item Name" label to "Hotkey name" for clarity
- Update documentation with detailed examples for single and multiple item usage

image

✅ this change compiles and runs
✅ this change has been tested
✅ this change does not break legacy behavior (incl. gwtoolbox.ini)
✅ this change does not break anything
✅ this change includes documentation


3vcloud:
Thanks, but why not use a chat alias to do this and more?

Le-Vagabond-gh:
I looked at the documentation for chat commands and "send command" hotkeys but I don't see a way to achieve this at all, could you explain?

as to why, I created a "use item" hotkey for my first miniature which worked great (the auto use on zoning is cool), and just wanted to use more than one (like I did on WoW).

3vcloud:
image

Le-Vagabond-gh:
I must be missing something, because I went through this section after your first reply:

as far as I can tell there's no chat command to "use item" (guildwars or toolbox) and even if there was the alias would just run all of them in sequence instead of picking a random one (resulting in no mini summoned because the second command dismisses it and then there's a global cooldown to summon a new one).

to get that triggered on zoning you would need a "send chat" hotkey as well, right? wouldn't having the functionality baked in the "use item" hotkey directly be interesting then, if only to simplify the user experience?

PR #1535, opened 2025-12-06
Improve RVA detection in injector exe

adituv:
Sometimes Windows doesn't clean up a previous Gw.exe process fully. When this happens, it's possible for the pid of the dead process to be lower, and for the RVA scans to fail for that process.

This patch loops over all candidate processes until it finds one where both character and email RVAs succeed, only failing if no processes have both scans succeed

### DIFF
--- a/GWToolbox/Inject.cpp
+++ b/GWToolbox/Inject.cpp
@@ -103,14 +103,24 @@ InjectReply InjectWindow::AskInjectProcess(Process* target_process)
         return InjectReply_NoProcess;
     }
 
-    uintptr_t charname_rva;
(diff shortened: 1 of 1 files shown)


adituv:
(Also C++ question: is there a way to write this with an iterator instead?  I couldn't figure out how to get a pointer out of it to pass to the ProcessScanner constructor)

henderkes:
> (Also C++ question: is there a way to write this with an iterator instead? I couldn't figure out how to get a pointer out of it to pass to the ProcessScanner constructor)

```cpp
for (const auto& process : processes);?
```

PR #1545, opened 2025-12-10
Improve error message for e.g. /tp zm when no nearest outpost found

adituv:
Previously, e.g. "/tp zm" when the character doesn't have access to the continent that the ZM is in would fail with "Your character does not have <current outpost> unlocked"


adituv:
`Chat::WriteChat` appends the terminating `\x1` for the \x107 unenc string parameter later so it ought to be fine

Actually, good point.  The `\x1` ought to crash it based on what I remember of the encstring decompilation.  It should maybe be `\x2\x108\x10a%s` instead so that the `\x1` terminates the `\x10a` encstring parameter instead

PR #1597, opened 2025-12-27
Feature/presearing daily quests

kyleshrader:
I've added a setting to the daily quest window which will show the presearing dailies while in a presearing region.

This change included adding encoded strings for the sandford trophies that were missing.

image

One weird bit of code was handling the subscriptions across items. Since items are repeated in the cycles, I have a check that compares before and after writing the daily to see if that frame included un/subscribing. If so, i go through and un/subscribe to all cycle indexes for that item. I preferred this approach to the one used in the Nick Traveller version, because that required a full special reimplementation of the write_daily_info function.

Please provide feedback where I could refactor things.

### DIFF
--- a/GWToolboxdll/Windows/DailyQuestsWindow.cpp
+++ b/GWToolboxdll/Windows/DailyQuestsWindow.cpp
@@ -108,7 +108,7 @@ namespace {
 
     static_assert(hard_coded_vanguard_names.size() == VANGUARD_COUNT);
 
-    std::unordered_map<DailyQuests::QuestData*, uint16_t> nicholas_sandford_item_collected_count;
+    std::unordered_map<std::wstring, uint16_t> nicholas_sandford_item_collected_count;
 
     DailyQuests::QuestData nicholas_sandford_cycles[] = {
         { MapID::None, GW::EncStrings::GrawlNecklaces },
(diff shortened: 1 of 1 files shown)


GWToolboxdll/Windows/DailyQuestsWindow.cpp
```diff
@@ -1106,40 +1125,53 @@ void DailyQuests::Draw(IDirect3DDevice9*)
...
-        ImGui::SameLine(offset += ws_width);
-    }
-    if (show_nicholas_in_window) {
-        ImGui::Text("Nicholas the Traveler");
-        ImGui::SameLine(offset += nicholas_width);
-    }
-    if (show_weekly_bonus_pve_in_window) {
-        ImGui::Text("Weekly Bonus PvE");
-        ImGui::SameLine(offset += wbe_width);
+    if (show_presearing) {
```
adituv: Is this the behaviour we want?  Do we want to swap out the contents in pre, or just make it available in the window?  It's already way too wide with all columns enabled so I'm not against this personally

kyleshrader: Showing all selected dailies would make it more customizable, including across characters pre and post. But less contextual.

GWToolboxdll/Windows/DailyQuestsWindow.cpp
```diff
@@ -108,68 +108,68 @@ namespace {
...
-        "Charr Carvings",
-        "Skale Fins",
-        "Dull Carapaces",
-        "Enchanted Lodestones",
-        "Charr Carvings",
-        "Spider Legs",
-        "Red Iris Flowers",
-        "Worn Belts",
-        "Dull Carapaces"
+    std::unordered_map<DailyQuests::QuestData*, uint16_t> nicholas_sandford_item_collected_count;
```
adituv: This works, but it could just be an array.  You're already looking at the index as the key for the cycles array below

kyleshrader: I've made it a map of the encoded item names, initially set the keys if they aren't set yet, and only check the inventory for the keys in the map. This reduces the checks from 52 to 13.

Issue #1626, opened 2026-01-08
Krak flamewhip - incorrect listed elite


Kherae:
https://wiki.guildwars.com/wiki/Krak_Flamewhip
currently TB lists him as having dwarven battle stance. he does not. he has devastating hammer.

PR #1628, opened 2026-01-08
fix(#1626): krak flamewhip skill

ChandlerFerry:


henderkes:
but thanks

ChandlerFerry:
Ah, apologies, noted for next time.

PR #1643, opened 2026-01-16
Account Inventory Window

ceddral0:
A searchable account wide inventory overview. Item stacks can be filtered and merged with more of the same item in the overview. Clicking on an item rerolls to the character/hero on which the item is. Ctrl+Click additionally moves the item to/from chest. Keeps track of free inventory spaces to better find the next not-yet-full storage character. Supports multiple accounts, if they share the same inventories directory.

Does anyone know a better way to track which hero an item belongs to? ATM it's based on loading order of heroes and items.

account_inventory_grid_view
account_inventory_detailed_view
account_inventory_free_slots

### DIFF
--- a/GWToolboxdll/Utils/TextUtils.h
+++ b/GWToolboxdll/Utils/TextUtils.h
@@ -19,7 +19,65 @@ namespace TextUtils {
     std::wstring Replace(const std::wstring_view subject, const std::wstring& pattern, const std::wstring& replacement);
     std::string Replace(const std::string_view subject, const std::string& pattern, const std::string& replacement);
 
-    std::string Base64Decode(std::string_view str);
+    template<typename CharT>
+    std::basic_string<CharT> Base64Decode(std::string_view encoded)
+    {
+        const std::string chars = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/";
+        std::basic_string<CharT> decoded;
+        std::vector<int> T(256, -1);
+
+        for (int i = 0; i < 64; i++)
+            T[chars[i]] = i;
+
+        int bits = sizeof(CharT) * 8;
+        int val = 0, valb = -bits;
+        for (unsigned char c : encoded) {
+            if (T[c] == -1) break;
+            val = (val << 6) + T[c];
+            valb += 6;
+            if (valb >= 0) {
+                decoded.push_back(CharT(val >> valb));
+                valb -= bits;
+            }
+        }
+        return decoded;
+    }
+
+    template<typename CharT>
+    std::string Base64Encode(std::basic_string_view<CharT> decoded)
+    {
+        const std::string chars = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/";
+        std::string encoded;
+        std::vector<int> T(64);
+
+        for (int i = 0; i < 64; i++)
+            T[i] = chars[i];
+
+        int bits = sizeof(CharT) * 8;
+        int val = 0, valb = -6;
+        for (CharT c : decoded) {
+            val = (val << bits) + c;
+            valb += bits;
+            while (valb >= 0) {
+                encoded.push_back(char(T[(val >> valb) & 0x3F]));
+                valb -= 6;
+            }
+        }
+        if (valb != -6) {
+            val = (val << 8);
+            valb += 8;
+            encoded.push_back(char(T[(val >> valb) & 0x3F]));
+            valb -= 6;
+        }
+        while (valb != -6) {
+            if (0 > valb && valb > -6) {
+                valb += 8;
+            }
+            encoded.push_back('=');
+            valb -= 6;
+        }
+        return encoded;
+    }
 
     std::string GetFormattedDateTime();
 
--- a/GWToolboxdll/Windows/AccountInventoryWindow.h
+++ b/GWToolboxdll/Windows/AccountInventoryWindow.h
@@ -48,20 +48,25 @@ class AccountInventoryWindow : public ToolboxWindow {
         uint32_t bag_id{};
         uint32_t slot{};
 
-        // further attributes necessary to display items
-        GW::Item item_partial{}; // everything needed to get a ShorthandItemDescription and GetItemImage
-        // memory management for allocated item_partial members
-        std::wstring info_string{};
-        std::wstring single_item_name{};
-        std::wstring complete_name_enc{};
-        std::wstring name_enc{};
-        std::wstring customized{};
+        uint32_t model_id{};
+        uint32_t model_file_id{};
+        uint32_t interaction{};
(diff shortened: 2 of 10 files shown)


GWToolboxdll/Utils/TextUtils.h
```diff
@@ -17,7 +17,65 @@ namespace TextUtils {
     std::string PrintFilename(std::string path);
     std::wstring PrintFilename(std::wstring path);
 
-    std::string Base64Decode(std::string_view str);
+    template<typename CharT>
+    std::basic_string<CharT> Base64Decode(std::string_view encoded)
+    {
+        const std::string chars = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/";
+        std::basic_string<CharT> decoded;
+        std::vector<int> T(256, -1);
```
henderkes: What is this?

GWToolboxdll/Windows/AccountInventoryWindow.h
```diff
@@ -0,0 +1,359 @@
...
+        bool operator()(const std::shared_ptr<InventoryItem> l, const std::shared_ptr<InventoryItem> r) const
+        {
+            return l->hero_id == r->hero_id && l->bag_id == r->bag_id && l->slot == r->slot && l->account == r->account && l->character == r->character;
+        }
+    };
+
+    struct MergeStack {
+        uint16_t quantity;
+        std::wstring description;
+        std::set<std::shared_ptr<InventoryItem>, ItemCompare> i;
```
henderkes: we most definitely don't want a std::set and we absolutely don't want shared_ptrs here

GWToolboxdll/Windows/AccountInventoryWindow.h
```diff
@@ -0,0 +1,359 @@
...
+            // fallback
+            if (delta == 0) delta = l->character.compare(r->character);
+            if (delta == 0) delta = l->location.compare(r->location);
+            if (delta == 0) delta = l->bag_id - r->bag_id;
+            if (delta == 0) delta = l->slot - r->slot;
+            if (delta == 0 && l->account < r->account) delta = -1;
+            return delta * sort_direction < 0;
+        }
+        // used for order inside MergeStack, i.e. for interaction and tooltip
+        bool operator()(const std::shared_ptr<InventoryItem> l, const std::shared_ptr<InventoryItem> r) const
```
henderkes: again

GWToolboxdll/Windows/RerollWindow.cpp
```diff
@@ -301,7 +304,11 @@ namespace {
         same_map = _same_map;
         same_party = _same_party;
         reroll_timeout = (reroll_stage_set = TIMER_INIT()) + 20000;
-        reroll_stage = PromptPendingLogout;
+        if (_do_not_prompt) {
+            reroll_stage = PendingLogout;
+        } else {
+            reroll_stage = PromptPendingLogout;
+        }
```
henderkes: reroll_stage = _do_not_prompt ? PendingLogout : PromptPendingLogout;

henderkes: But why is this logic touched at all?

GWToolboxdll/Windows/AccountInventoryWindow.h
```diff
@@ -0,0 +1,359 @@
...
+        std::wstring info_string{};
+        std::wstring single_item_name{};
+        std::wstring complete_name_enc{};
+        std::wstring name_enc{};
+        std::wstring customized{};
+
+        // caches, do not serialize
+        IDirect3DTexture9** texture; // output of GetItemImage
+        std::wstring description{}; // output of AsyncDecodeStr(ShorthandItemDescription)
+        std::string location{}; // (Player) <Storage Pane> or <Hero Name>
```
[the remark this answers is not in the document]
henderkes: Makes sense if it comes from Imgui, but why does it come from Imgui?

3vcloud:
Rather than making reviews like the above, I usually just merge it in and then fix it myself when I have time

GWToolboxdll/Windows/AccountInventoryWindow.h
```diff
@@ -0,0 +1,359 @@
...
+
+    };
+
+    struct MergeStack;
+    struct ItemCompare {
+        ImGuiTableSortSpecs* sort_specs{};
+        std::wstring current_account{};
+        // used for order in Draw
+        bool operator()(const std::shared_ptr<MergeStack> lms, const std::shared_ptr<MergeStack> rms) const
```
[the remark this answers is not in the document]
ceddral0: The problem I was originally using shared_ptrs for was, to handle decoded descriptions coming from AsyncDecodeStr. But looking at it now, I just misunderstood how to use AsyncDecodeStr. Thank you, I'll get rid of all the shared_ptrs. Might take a bit to get them all out of there :)

GWToolboxdll/Windows/AccountInventoryWindow.h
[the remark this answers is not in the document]
ceddral0: I'm not sure I understand the question. I have an input field so users can filter their items by a hero name, for example to show all items that sit on a Koss. Is there something other than ImGui::InputText I'm supposed to use for that?

GWToolboxdll/Windows/RerollWindow.cpp
ceddral0: The gather inventories function rolls through all characters and heroes to gather all inventory contents. It checks beforehand if any of the chars are in an outpost where toolbox is disabled and thus the rerolling would stop. If that's the case it shows the user a little note, that they should move that character, including a button to reroll to it. If this button does not skip the 'internal' reroll prompt there'd be 2 popups. I could either accept that there are 2, or remove my own popup. But I wanted to show a more specific note telling the user why they have do something before 'gather inventories' can work.

That said, I'd be fine without the specific popup note, in which case this logic change isn't necessary.

ceddral0:
> Rather than making reviews like the above, I usually just merge it in and then fix it myself when I have time

I think I'd like to at least fix the shared_ptrs, before you take your time with this. Instead of making you comb through all of those.

Somewhat unrelated, I'm not sure why this pullrequest sits one commit below my dev branch head after I rebased it onto upstream dev. But perhaps that will fix itself when I push a new commit removing the shared_ptrs.

3vcloud:
> > Rather than making reviews like the above, I usually just merge it in and then fix it myself when I have time
> 
> I think I'd like to at least fix the shared_ptrs, before you take your time with this. Instead of making you comb through all of those.
> 
> Somewhat unrelated, I'm not sure why this pullrequest sits one commit below my dev branch head after I rebased it onto upstream dev. But perhaps that will fix itself when I push a new commit removing the shared_ptrs.

Ok, I won't merge your code in until you've sorted that. I wanted to get a new version out this weekend, but this feature might not make it in for then

ceddral0:
Finally got around to working on this again.

PR #1644, opened 2026-01-18
mission Variable for EliteBossLocation and More

Asinh17:
- Adds a `mission` variable to the `EliteBossLocation` struct
  - Further standardize how mission bosses are marked
  - Removes now redundant information
- Adds a few more missing bosses and fixes small errors
  - "Norn Catering" quest
  - "Dajkah Inlet" challenge mission
  - "Kodonur Crossroads" mission
- Reorders a few bosses to improve visibility on the map


henderkes:
Good god, what a list. Bless you Jon.

Asinh17:
The diff looks way bigger than it is. In most cases it is just a single `false` added to the line.
I was wondering if it would make sense to move the EliteBossLocation const into an SQLite database or at least a JSON since it's essentially harcoded data. I would be happy to help with that, if it makes sense.

Asinh17:
After converting to JSON, I realized that there is an issue with it. No autocomplete for skill and map ids, since I have to store them as strings, might make make maintaining this more error prone.
Let me know if I should still commit the JSON or keep the PR as is.

henderkes:
keep as is

3vcloud:
Roll everything back; use GW::Map::GetMapInfo(boss->map_id) to check if the map type is mission.

PR #1710, opened 2026-02-08
Feature(#1252): Add minimap snapping to mission map

efrenmanuel:
This PR aims to add support to have the minimap be snapped to the mission map. 
Sort of an "improved"/extended compass.

Status:

- [X] Map snaps to Mission Map
- [X] Map is set to the same size as Mission Map
- [x] Compass related options get disabled when the  snap to Mission Map option is enabled
- [x] Map updates when the Mission Map is modified
  - This is not ideal, I couldn't find the hook for when the Mission Map is moved, so the possition update is forced with every update
- [x] Map scales with the Mission Map.
  - This is done with a magic number in the way, it would be nice to know why that number works.
- [x] Map position matches the Mission Map's.

Cleanup:
- [x] Remove TODOs
- [x] clang-format
- [x] clang-tidy 

Fixes:
-  [x] Rotation needs to be disabled when its snapped to mission map
-  [x] Not disable quest markers in the ui callback
-  [x] When the minimap is snapped to compass again, fix zoom

Edit:
I just found out that this resolves https://github.com/gwdevhub/GWToolboxpp/issues/1252

### DIFF
--- a/GWToolboxdll/Widgets/Minimap/Minimap.cpp
+++ b/GWToolboxdll/Widgets/Minimap/Minimap.cpp
@@ -1026,7 +1026,7 @@ void Minimap::DrawSettingsInternal()
         if (ImGui::Checkbox("Compass flagging controls", &hide_flagging_controls)) {
             hide_flagging_controls_patch.TogglePatch(hide_flagging_controls);
         }
-        ImGui::ShowHelp("Takes effect on map change. Doesn't work in PvP as Toolbox is disabled there."); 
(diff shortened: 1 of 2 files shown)


3vcloud:
Actually I'll pick this up and run with it, its easier to amend on my side

efrenmanuel:
Hmm interesting, does it work with the float for you?
I wonder if it may be because of my monitor's 21:9 aspect ratio.
I tried it at different ui scales but haven't tried with different resolutions/aspect ratios.

PR #1729, opened 2026-02-17
Add damage and heal tracker to Observer Module

taitai42:
This allows to track damage and heal done in Observer mode, in both Observer party window, and Observer Player for advanced metrics. (new fields in red squares red) 

image

It's a bit funky as we don't have the max hp of a target available at all time (only when we obs the player, and it seems like it's only refreshed when the player health is updated (by dmg or heal)) and the damage packet are sent as % of max hp. 
It means we have to calculate the actual damage done every time a target receives damage, and therefore have this max hp value. 
I've made a sort of cache system to keep a value for each player in the game, that would be updated if you obs it, and I've set a default value of 530hp for everybody, that should be close to what players currently runs on an average of every position. 
If someone has a better way, I'm all ears. 

Also added it to the json export module so Arka can import it to gwrank. 

### Player Observer

- Total damage made & received
- Total heal made & received
- heal made & received to individual allies
- damage made & received to individual opponents
- damage made by individual skills

### Party Observer

- total damage made & received
- total heal made & received
- max hp detected

### Json exporter

- all above metrics
- damage made to parties (without npc)

## BugFix
- Now correctly detects the winning team and moral boosts.

### DIFF
--- a/GWToolboxdll/Modules/ObserverModule.cpp
+++ b/GWToolboxdll/Modules/ObserverModule.cpp
@@ -8,11 +8,14 @@
 #include <GWCA/GameEntities/Player.h>
 #include <GWCA/GameEntities/Skill.h>
 
+#include <GWCA/Packets/StoC.h>
+
 #include <GWCA/Managers/MapMgr.h>
 #include <GWCA/Managers/SkillbarMgr.h>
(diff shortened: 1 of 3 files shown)


BearsMan:
The target branch should be dev, not master.

PR #1758, opened 2026-03-05
feat/minimap: marked targets inherit custom size/shape

gam415:
### Use-case

1. I defined custom color/size agents for Skeletons in UW (size: `120.0`), and Coldfires (size: `default`).
2. I use `/marktarget` command on a patrolling Skeleton to keep track of it. Now it has the default marked size (`75.0`) and is actually *less* visible. I wanted its size to remain at `120.0`, just update its color to distinguish it from other skeles.
3. If I update "marked agents size" to `120.0` to fix that, then when I mark a Coldfire it will now be of size `120.0` too (I wanted it to remain at `default`, and only update its color).

### Solution

We introduce a new checkbox setting which makes marked targets inherit the custom size/shape of custom agents if any is defined. I believe this is more aligned with what users would want out of the `/marktarget` feature, but for backwards-compatibility the setting is disabled by default.

### Alternative

An alternative would be to always inherit custom agent sizes when `marked_agent_size == default_size`. And if a user updated the `marked_agent_size` to something other than `default_size`, we could assume they chose to override the marked agent size, and use it instead. 

Since this behavior is not backwards-compatible, I did not implement it.

### DIFF
--- a/GWToolboxdll/Widgets/Minimap/AgentRenderer.cpp
+++ b/GWToolboxdll/Widgets/Minimap/AgentRenderer.cpp
@@ -207,6 +207,7 @@ void AgentRenderer::LoadSettings(const ToolboxIni* ini, const char* section)
     LoadDefaultSizes();
     
     LOAD_FLOAT(size_marked_target);
+    LOAD_BOOL(marked_target_inherit_custom_agents);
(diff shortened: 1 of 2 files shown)


henderkes (approved):
Thank you!

PR #1759, opened 2026-03-06
Update KeyboardLanguageFix.cpp

henderkes:


henderkes:
In case the game has already loaded and the user switched back to their language, we don't want to switch them to en-US again.

PR #1766, opened 2026-03-09
Minimap extensibility

awgil:
A bunch of changes to support extending minimap from plugins.

Note that this PR includes some more changes - I've tried to split them into separate commits with reasonable descriptions. Technically only last commit implements minimap extensibility - the rest is optional; the only thing last commit relies on is dllexport machinery introduced in 'Logging improvements' commit.

If desired, I can split that logging commit further into two (dllexport and logger markup). Or I can move some changes out into separate PRs (or even remove them if they are not desired)...

### DIFF
--- a/Core/CMakeLists.txt
+++ b/Core/CMakeLists.txt
@@ -1,5 +1,5 @@
 
-FILE(GLOB SOURCES
+FILE(GLOB SOURCES CONFIGURE_DEPENDS
     "*.h"
     "*.cpp")
 
--- a/GWToolbox/CMakeLists.txt
+++ b/GWToolbox/CMakeLists.txt
@@ -1,5 +1,5 @@
 
-FILE(GLOB SOURCES
+FILE(GLOB SOURCES CONFIGURE_DEPENDS
     "*.h"
     "*.cpp"
     "GWToolbox.rc")
--- a/GWToolboxdll/CMakeLists.txt
+++ b/GWToolboxdll/CMakeLists.txt
@@ -1,6 +1,6 @@
 add_library(GWToolboxdll SHARED)
 
-file(GLOB SOURCES
+file(GLOB SOURCES CONFIGURE_DEPENDS
     "*.h"
     "*.cpp"
     "GWToolbox.rc"
@@ -105,10 +105,10 @@ if(NOT FXC)
     message(SEND_ERROR "unable to find Fxc.exe for HLSL compilation")
 endif(NOT FXC)
 
-file(GLOB HLSL_VERTEX_SHADER_FILES
+file(GLOB HLSL_VERTEX_SHADER_FILES CONFIGURE_DEPENDS
     Widgets/Minimap/Shaders/*_vs.hlsl
 )
-file(GLOB HLSL_PIXEL_SHADER_FILES
+file(GLOB HLSL_PIXEL_SHADER_FILES CONFIGURE_DEPENDS
     Widgets/Minimap/Shaders/*_ps.hlsl
 )
 
--- a/GWToolboxdll/CMakeLists.txt
+++ b/GWToolboxdll/CMakeLists.txt
@@ -24,8 +24,18 @@ file(GLOB SOURCES
      
 source_group(TREE "${CMAKE_CURRENT_SOURCE_DIR}" FILES ${SOURCES})
 target_sources(GWToolboxdll PRIVATE ${SOURCES})
- 
+
+# auto-generated files - export header & pch 
+include(GenerateExportHeader)
+generate_export_header(GWToolboxdll)
+#target_sources(GWToolboxdll PRIVATE FILE_SET generated_header TYPE HEADERS BASE_DIRS ${CMAKE_CURRENT_BINARY_DIR} FILES ${CMAKE_CURRENT_BINARY_DIR}/gwtoolboxdll_export.h)
+target_sources(GWToolboxdll PRIVATE ${CMAKE_CURRENT_BINARY_DIR}/gwtoolboxdll_export.h)
+target_include_directories(GWToolboxdll PUBLIC ${CMAKE_CURRENT_BINARY_DIR})
+source_group(Generated FILES ${CMAKE_CURRENT_BINARY_DIR}/gwtoolboxdll_export.h)
+
 target_precompile_headers(GWToolboxdll PRIVATE "stdafx.h")
+source_group(Generated REGULAR_EXPRESSION cmake_pch)
+
 target_compile_definitions(GWToolboxdll PRIVATE
     "_USRDLL"
     "GWCA_IMPORT"
@@ -88,7 +98,7 @@ target_link_libraries(GWToolboxdll PRIVATE
     Dbghelp.lib # for MiniDump
 	delayimp # for delay loading gwca
     )
-      
+
 file(GLOB FXC_PATHS "C:/Program Files (x86)/Windows Kits/10/bin/*/x86")
 find_program(FXC fxc DOC "hlsl compiler" PATHS ${FXC_PATHS})
 if(NOT FXC)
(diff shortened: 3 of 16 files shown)


henderkes:
A bunch of good changes here, especially the cmake ones. I suppose the minimap registration logic needs to be tested with multiple plugins loading and unloading in different orders, but it also looks good.

Only questions I have:
- why export logging functions? plugins can already log to console/chat themselves
- I don't think the minor performance impact of localtime matters, but it's fine anyway. If someone gets unexpected timestamps twice a year so be it

3vcloud:
> A bunch of good changes here, especially the cmake ones. I suppose the minimap registration logic needs to be tested with multiple plugins loading and unloading in different orders, but it also looks good.
> 
> Only questions I have:
> 
> * why export logging functions? plugins can already log to console/chat themselves
> * I don't think the minor performance impact of localtime matters, but it's fine anyway. If someone gets unexpected timestamps twice a year so be it

I don't think I've got an issue with the logging - it helps to reduce the footprint/code debt at any rate by just exporting the funcs.

The renderer registration is good aswell, and because we've wrapped the details up in the minimap context object now, you shouldn't need anything more complicated than that to do what you need to do

The only bad thing about this is that it gives plugins more blame for crashing, but we always tell people to test without plugins before reporting a crash anyway so all good

henderkes:
Yeah I don't think it's an issue either, but it's such a minimal footprint that plugins got from header includes either way.

awgil:
> why export logging functions? plugins can already log to console/chat themselves

How? I couldn't find a way to do this... Note that normal 'printf' type calls don't work - if I understand correctly, since we're linking CRT statically, each dll (toolbox and each plugin) will have its own version of stdout etc - and only toolbox one will be redirected to file. I've asked in discord and people have told me they roll out their own logging to own file in their plugins, which doesn't look great imo.

Now, initially I've only exported functions dealing with logging to file. However, then I found that something (assert macro iirc) also uses FatalError, so it has to be exported too. At this point I thought it makes sense to just export the chat functions too.

> I don't think the minor performance impact of localtime matters, but it's fine anyway. If someone gets unexpected timestamps twice a year so be it

So the main thing here is - I wanted millisecond precision in timestamps - and C timing utilities (time_t typedef, time(), tm struct) have second precision. Now, chrono's to_local does a timezone db lookup every call (for reasons described in comments), *unlike* C's localtime (it just caches timezone info on init). There are quite a few complaints about the performance effect it causes on the internet, and I was really wary about not introducing perf regressions in something as (potentially) high frequency as a logging function. So in effect, with the way I did it we preserve the old behaviour on timezone changes (monotonous time rather than timezone correct). Happy to change it though.

henderkes:
> > why export logging functions? plugins can already log to console/chat themselves
> 
> How? I couldn't find a way to do this... Note that normal 'printf' type calls don't work - if I understand correctly, since we're linking CRT statically, each dll (toolbox and each plugin) will have its own version of stdout etc - and only toolbox one will be redirected to file. I've asked in discord and people have told me they roll out their own logging to own file in their plugins, which doesn't look great imo.

Just by redirecting stdout/stdderr to the active console. And no, whether linking the CRT statically or dynamically doesn't make a difference here. stdout/in/err are OS file descriptors and not related to the runtime linkage.

>So the main thing here is - I wanted millisecond precision in timestamps - and C timing utilities (time_t typedef, time(), tm struct) have second precision. Now, chrono's to_local does a timezone db lookup every call (for reasons described in comments), unlike C's localtime (it just caches timezone info on init). There are quite a few complaints about the performance effect it causes on the internet, and I was really wary about not introducing perf regressions in something as (potentially) high frequency as a logging function. So in effect, with the way I did it we preserve the old behaviour on timezone changes (monotonous time rather than timezone correct). Happy to change it though.

Since stdout logging is only active in debug builds and should be sparingly used in the game chat, I don't think that really matters. If it helps your use case, no worries, I was just curious.

awgil:
> Just by redirecting stdout/stdderr to the active console

The console is available only in debug builds (I haven't actually checked how it works there) - I was thinking more about non-debug (ie RelWithDebInfo) builds, where logging is redirected to a file. Just calling printf from a plugin does not write the output to the log file.

> stdout/in/err are OS file descriptors and not related to the runtime linkage

Are you sure? I can investigate more, but from what I've seen stepping through CRT code while trying to figure out why printf wasn't logging as expected - I got the impression stdout etc is CRT concepts (static FILE objects) rather than OS concepts...

> Since stdout logging is only active in debug builds

Stdout logging is definitely active in release builds, logging out to file (and personally I prefer logging to file way more than to console, but that's ofc personal preference).

awgil:
I did a bit more testing on a simple project (two dlls, each having an 'init' function that reopens stdout to own file and 'print' function that writes a message; then executable that calls initA -> initB -> printA -> print B).

So in case of dynamic CRT linking, file A is empty, file B contains msgA followed by msgB (as expected).

In case of static CRT linking, file A is also empty (so yeah apparently even though each dll now has its own FILE structure for stdout, some underlying file handle is shared), but file B now contains msgB followed by msgA (in reverse order).

So maybe the problem is with some CRT level buffering instead, which is independent for two copies of CRT?..

3vcloud:
This is merged including logging, dw about it

awgil:
> This is merged including logging, dw about it

I know :) I'm just really interested in getting to the bottom of this CRT handle thing, I haven't dealt with statically linking CRT before much.

Sooo having debugged a bit more - I'm now pretty sure that everything stdout etc related is independent in every dll when CRT is linked statically. When you do a freopen(stdout) in libA, it would go through a few layers - there's a static array of three handles stdin/out/err (which are duplicated in each dll), this gets mapped to a file handle (independently for each dll), which would then be used to ultimately lookup win32 HANDLE in a static array.

So when initA is called, it maps libA!_iob[1] to CRT file handle 1, which gets mapped to win32 HANDLE 0xC. It would then close it, and then later open another file - which turns out to reuse win32 handle 0xC (but I don't think there's anything to guarantee that opening a file immediately after closing a handle reuses the same handle id!).

At the same time, the copy of CRT structures in libB still contains win32 handle 0xC in its mapping tables - so it's 'just luck' that it still points to a valid file.

Anyway, conclusion is, it's generally unsafe to use stdout and friends in plugins - stdout might or might not contain a valid win32 handle after toolbox dll calls freopen. It's kind of a 'use-after-free' situation. Worse, if you do freopen in a plugin, then it might fuck up with CRT mappings in main toolbox dll...

henderkes:
> Anyway, conclusion is, it's generally unsafe to use stdout and friends in plugins - stdout might or might not contain a valid win32 handle after toolbox dll calls freopen. It's kind of a 'use-after-free' situation. Worse, if you do freopen in a plugin, then it might fuck up with CRT mappings in main toolbox dll...

No, there's no luck involved there, you just need to bind the stdout/stderr to the existing console window, which you can get with the winapi. I'm not sure if this is race free if multiple dll's want to log at the same time, but we've been doing it fine with no issues forever. gMod, daybreaks dll and toolbox are all statically linked and write to the same console in debug mode.

awgil:
I think this all relies on the fact that when you're calling freopen, ultimately there's a sequence of CloseHandle -> CreateFile, and CreateFile happens to reuse the same handle value (which is in turn a global per-process thing). Technically I think it's not (and even can't!) be guaranteed - eg imagine some other thread creates some other handle concurrently, and happens to execute between CloseHandle and CreateFile - hence the 'luck'.

The win32 handle value is then stored in multiple different places (eg in `__pioinfo` array in every copy of CRT, and also passed to SetStdHandle - which is how these different CRT copies originally get the same handles for stdout on init!).

Also apparently the buffering is also done on CRT level, which means that different copies of CRT logging to the 'same' stdout will be buffered independently, which can make messages interleaved (again, don't know how it works when stdout is redirected to allocated console in debug build, maybe there's no buffering then).

gam415:
Is there anything in here that will break plugins that already render on the minimap...? Hoping not.

PR #1772, opened 2026-03-14
Make Color::Save/Load usable in plugins.

awgil:
ToolboxPlugin::ini is CSimpleIni rather than derived ToolboxIni class; distinction is not relevant for Color serialization functions, but it taking pointer to derived class prevents it being used normally in plugins.

### DIFF
--- a/GWToolboxdll/Color.h
+++ b/GWToolboxdll/Color.h
@@ -57,7 +57,7 @@ namespace Colors {
     static constexpr Color BlueGrey() { return RGB(96, 125, 139); }
     static constexpr Color Grey() { return RGB(158, 158, 158); }
 
-    static Color Load(const ToolboxIni* ini, const char* section, const char* key, const Color def)
+    static Color Load(const CSimpleIni* ini, const char* section, const char* key, const Color def)
     {
         auto ParseUInt = [](const char* str, unsigned int* val, const int base)
         {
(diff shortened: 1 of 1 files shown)


henderkes:
thanks

3vcloud:
Hasn't this now caused you to not know where the ini is being stored? I'm sure I read conversation yesterday asking how we know where a plugin's ini is located inside Load/Save functions, and the answer to that would have been "ToolboxIni includes this info"

henderkes:
> Hasn't this now caused you to not know where the ini is being stored? I'm sure I read conversation yesterday asking how we know where a plugin's ini is located inside Load/Save functions, and the answer to that would have been "ToolboxIni includes this info"

Oh gosh, you're right. CSimpleIni doesn't have the path.

henderkes:
Oops, it doesn't need the path there as it only interfaces with the already loaded ini. However, plugins should use ToolboxIni instead of CSimpleIni.

3vcloud:
No, I think the reason for asking was to indirectly find a suitable place to save a separate ini file that would be done by finding the parent directory of the file name provided via the toolboxini struct an using that

awgil:
> Hasn't this now caused you to not know where the ini is being stored? I'm sure I read conversation yesterday asking how we know where a plugin's ini is located inside Load/Save functions, and the answer to that would have been "ToolboxIni includes this info"

So, this PR only changed serialization functions to accept CSimpleIni* (base class) rather than ToolboxIni, because former is what plugins have, and Color serialization doesn't need anything derived class provides.

Now, you might want to change the plugin to store derived class instead (and export it etc), this would be fine too. However, it wouldn't actually resolve anything about the issue I was talking about in discord (the first moment the settings path is available is LoadSettings) - I can just as well just store it in a field once it's available, it would be functionally equivalent to what one can do with ToolboxIni.

PR #1775, opened 2026-03-15
Add healing tracking to damage meter

gunnarbeutner:
This adds healing stats to `/damage` and the damage widget. Like damage stats this does not account for indirect "healing", e.g. through regeneration or damage mitigation.

image

image


3vcloud:
I was just thinking about this on a playthrough this morning, thanks!

PR #1778, opened 2026-03-21
VQ mission map overlay

gunnarbeutner:
This adds a button to the mission map which lets you turn on an overlay that has VQ-specific features:

1. It shades the walkable area differently from inaccessible parts of the map.
2. As you explore the map the explored parts are shaded differently (well, transparent). This is on a per-session basis.
3. There's a circle around your current position (based on radar range). The boundary between explored and unexplored space is shown in yellow.
4. Enemy positions are tracked as you encounter them, even when leaving radar range.
5. When you return to a part of the map where you encountered enemies before but they're no longer there an orange icon with an arrow pointing in the direction the enemy was last seen moving towards (for patrols).
6. The bottom left corner of the mission map shows how many enemies we know about and how many enemies remain in total.
7. Context menu item that automatically sets a quest marker on the nearest enemy (auto-updates when you move or the enemy dies).
8. Spirits and minions are ignored.

https://github.com/user-attachments/assets/b8d23964-50f9-4216-9114-f0f34bd4887d

### DIFF
--- a/GWToolboxdll/Widgets/MissionMapWidget.cpp
+++ b/GWToolboxdll/Widgets/MissionMapWidget.cpp
@@ -200,6 +200,7 @@ namespace {
         };
 
         std::vector<Vertex> vertices;
+        constexpr float LINE_HALF_THICKNESS = 1.5f; // 3px wide route lines
 
         for (const auto& line : lines) {
             if (!line->visible) continue;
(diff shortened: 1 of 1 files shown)


henderkes:
very nice addition, I will review later unless jon beats me to it

3vcloud:
this tanks fps so I'll take a look at it, but looks really cool!

gunnarbeutner:
Hmm, I haven't noticed that, but will have a closer look this evening or tomorrow. Been playing via Steam Remote Play those past few days so FPS was the least of my worries.

3vcloud:
so the issue is more visible in debug mode due to all the garbage debug checks that the std namespace likes to add, but it highlighted to me that we're calculating positions/coords in the CPU, creating a massive list of vectors manually every frame before giving it to the device.

The minimap draw terrain once in a buffer, and we then use an ortho projection, cropping and translating to move the map around within the context of the minimap area on screen, which is much less expensive and we don't have to rebuild the vertices because we dont care where they are individually on-screen at that point

I'll see if I can refactor it myself to do this

henderkes:
Let me know if you need me, I'm free tomorrow

gunnarbeutner:
See #1779. With the fixes I'm not seeing any FPS drops anymore. I was previously testing this with RelWithDebInfo where that wasn't really a problem at all.

PR #1779, opened 2026-03-21
Performance improvements for the VQ map

gunnarbeutner:
This should fix some the performance issues introduced by the VQ map overlay. I've been testing this with RelWithDebInfo before and hadn't noticed how bad it was with the Debug build.

This also adds anti-aliasing for the radar range overlay and smooths out the map border a bit.

### DIFF
--- a/GWToolboxdll/Widgets/MissionMapWidget.cpp
+++ b/GWToolboxdll/Widgets/MissionMapWidget.cpp
@@ -348,16 +348,16 @@ namespace {
         {
             constexpr DWORD INACCESSIBLE_COLOR = D3DCOLOR_ARGB(190, 0, 0, 0);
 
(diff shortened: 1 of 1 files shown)


3vcloud:
Thanks, but as I mentioned I was going to tackle the fps issues so there are now merge conflicts :(

gunnarbeutner:
No worries, I'll just drop the FPS fix and rebase the rest. The anti-aliasing stuff should still be useful though. :)

3vcloud:
ok sounds good. Might be worth moving some of this logic outside of the 1 source file to avoid these issues later

gunnarbeutner:
Well, it's not being used elsewhere - so I've kept it in just the one file for now. The updated patch should cut down the number of CPU projections for the cells significantly. Try and see if performance is any better.

PR #1780, opened 2026-03-21
Fix VQ fog rendering in outposts

gunnarbeutner:
When returning to an outpost when the VQ overlay was previously active the mission map would have the "inaccessible area" shading everywhere. Oops.


3vcloud:
did you test this? all the counts are set to 0 and are only reset to something when is_explorable is true, so this isnt needed

gunnarbeutner:
Yeah, in fact I can reproduce this by just walking into an explorable and then back into an outpost: https://youtu.be/KswtfmU751k

gunnarbeutner:
I've retested this against the latest dev branch and found a few (other) issues:

1. The radar range circle wasn't being rendered.
2. When joining an explorable area for the very first time after a restart the entire map was shaded as if it was explored.
3. The area of the world map outside of the current map wasn't being rendered with the "inaccessible" shade.

PR #1781, opened 2026-03-22
Add silent flag to /hero command

gunnarbeutner:
`/hero guard silent`, `/hero attack Koss silent`, etc. suppress the hero behavior chat messages for 1 second after the command is issued.


3vcloud:
Make the ChatSettings thing more generic, similar to `AudioSettings::BlockSoundForMs(const wchar_t* filename, clock_t ms)` - you'd call it 3 times, once for each behaviour enc string. We can then use it for other features easily later when we need similar functionality. It'll basically be a temporary chat filter

PR #1786, opened 2026-03-22
Fix VQ overlay tracking pets and allegiance-changed NPCs

### LINKED ISSUE #1785: Norn bounty npcs showing up as stale enemies on vq overlay
check alliegiance in the code; if its no longer enemy, remove it

gunnarbeutner:
Remove tracked enemies when their allegiance changes away from enemy. Also exclude pets from enemy tracking.

Fixes #1785

### DIFF
--- a/GWToolboxdll/Widgets/MissionMapWidget.cpp
+++ b/GWToolboxdll/Widgets/MissionMapWidget.cpp
@@ -496,7 +496,11 @@ namespace {
             if (!agent) continue;
             const auto living = agent->GetAsAgentLiving();
             if (!living) continue;
-            if (living->allegiance != GW::Constants::Allegiance::Enemy) continue;
+            if (living->allegiance != GW::Constants::Allegiance::Enemy) {
+                // Allegiance changed — stop tracking
+                tracked_enemies.erase(agent->agent_id);
+                continue;
+            }
             const auto npc = GW::Agents::GetNPCByID(living->player_number);
             if (npc && (npc->IsSpirit() || npc->IsMinion())) continue;
 


GWToolboxdll/Widgets/MissionMapWidget.cpp
```diff
@@ -496,9 +496,13 @@ namespace {
...
             if (!living) continue;
-            if (living->allegiance != GW::Constants::Allegiance::Enemy) continue;
+            if (living->allegiance != GW::Constants::Allegiance::Enemy) {
+                // Allegiance changed — stop tracking
+                tracked_enemies.erase(agent->agent_id);
+                continue;
+            }
             const auto npc = GW::Agents::GetNPCByID(living->player_number);
-            if (npc && (npc->IsSpirit() || npc->IsMinion())) continue;
+            if (npc && (npc->IsSpirit() || npc->IsMinion() || npc->IsPet())) continue;
```
henderkes: Pets can be required for a vanquish once aggroed, so I don't tgink we can skip them here.

gunnarbeutner: It was my understanding that pets don't count (as opposed to unowned animals which do count once aggro'd). But somehow IsPet() is true for non-pet/non-animal enemies. So I've removed the check for now.

henderkes: Yeah they count as pets too :P

henderkes:
uhm what happened here, did you accidentally change base to some ancient branch?

3vcloud:
Looks like another feature was accidentally merged in

gunnarbeutner:
Oops. Let me fix that.

PR #1790, opened 2026-03-24
Fix custom renderer lines not drawing on mission map when VQ overlay …

gunnarbeutner:
…is disabled

SubmitVertexBuffers was called inside the `if (show_vq_overlay)` block in DrawExplorable, so custom renderer lines (e.g. Place Marker routes) were never submitted when VQ was off. Move the call outside the VQ block and gate only VQ-specific geometry (borders, fog, enemies) behind the toggle.


henderkes:
Thanks!

3vcloud:
sorry I clobbered this, I had refactored a shitload of code before work and didn't have a chance to commit - it fixed this issue too

PR #1795, opened 2026-03-29
Fix VQ overlay immediately marking enemies as stale when leaving radar

gunnarbeutner:
Enemies now stay marked as alive when they leave radar range. They only transition to stale when you re-enter range of their last known position and they're no longer there.


gunnarbeutner:
The two other commits fix two more issues I came across:

* The world-to-screen transform didn't work on some maps, including DoA and Gate of Madness (probably other missions in the torment region as well, but haven't checked).
* The VQ button was visible in outposts but had no real effect there.

PR #1801, opened 2026-03-30
Fix VQ overlay flicker when mission map pans

gunnarbeutner:
When moving around both the radar circle and the map boundary outline would "wobble". This should fix that (or at least noticeably reduce the problem).

Also, it makes the exploration cells a bit smaller and re-introduces smoothing so the map boundary is less jagged:

Before:
image

After:
image


3vcloud:
The poly line stuff seems like a lot of extra code just for smoothing, I'm not sure if it's worth that?

Good catch on the rounding flicker and the g2s tweaks though

gunnarbeutner:
Nah, I guess not. With the smaller cell size it's barely noticeable.

PR #1811, opened 2026-04-02
fix: replace <> with [] in error messages

gam415:
I'm not really sure what is causing it, but it seems that using the `<arg>` syntax in these messages is causing GW to interpret it in some special way. I know the convention is `<>` for required arguments, but switching to `[]` fixes the issue, even though it usually means the argument is optional. Maybe the problem can be fixed differently I don't know.

Before with `/useskill` (no arg):

<details>
useskill_error_1
</details>

After with `/useskill`, `/pc`, `/withdraw` (no arg):

<details>
useskill_error_2
</details>

### DIFF
--- a/GWToolboxdll/Modules/ChatCommands.cpp
+++ b/GWToolboxdll/Modules/ChatCommands.cpp
@@ -661,13 +661,13 @@ namespace {
     }
 
 
-    constexpr auto withdraw_syntax = "'/withdraw <quantity (1-65535)> [model_id1 model_id2 ...]' tops up your inventory "
+    constexpr auto withdraw_syntax = "'/withdraw [quantity (1-65535)] [model_id1 model_id2 ...]' tops up your inventory "
         "with a minimum quantity of 1 or more items, identified by model_id\n"
-        "If no model_ids are passed, withdraws <quantity>[k] gold from storage\n"
+        "If no model_ids are passed, withdraws [quantity][k] gold from storage\n"
         "If quantity is 'all' and you do not pass model_ids, withdraws all gold you have or can hold.";
-    constexpr auto deposit_syntax = "'/deposit <quantity (1-65535)> [model_id1 model_id2 ...]' deposits <quantity> items, "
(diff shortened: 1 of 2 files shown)


henderkes (changes requested):
that's incorrect though, `[]` arguments are optional, `<>` arguments are not

gam415:
Also a nitpick: the warning message for `/useskill` (no arg) explicitly instructs the user that to stop a skill, they may type `/useskill` or `/useskill stop`, however `/useskill` triggers the warning message, so it doesn't seem to be what we want the user to type. 

Should the warning message say `/useskill 0` or `/useskill stop`? These are the arguments that don't trigger the warning.

gam415:
@henderkes Yes obviously, as stated in the PR. I'm just raising the problem, if yall know a better way to fix it, I'm all ears.

henderkes:
Ah sorry, the notification took me straight to the files changed tab, so I didn't read the comment. Not sure how to fix it yet, there'll likely be an escape character to make <> work.

henderkes:
> Also a nitpick: the warning message for `/useskill` (no arg) explicitly instructs the user that to stop a skill, they may type `/useskill` or `/useskill stop`, however `/useskill` triggers the warning message, so it doesn't seem to be what we want the user to type.
> 
> Should the warning message say `/useskill 0` or `/useskill stop`? These are the arguments that don't trigger the warning.

If /useskill acts the same as `/useskill stop` (i.e. it stops execution) then the error message shouldn't come. If it results in an error and doesn't stop execution, the instruction needs to be updated.

3vcloud:
Not a hill to die on - thanks for the fix

gam415:
It seems that other commands in tb already break the <> as mandatory arg convention, `/config` does: `/config set|get|toggle| [section key [value]]` (I was just thinking about that command before the merge and hoping there was some precedent so that we can stay consistent with it). I think I'll still fix the `/useskill` warning thing too.

PR #1812, opened 2026-04-02
Update '/useskill' syntax message for current behavior

gam415:
It seems parts of this message were relevant when multiple `/useskill` were permitted (i.e. "to stop all skills"). Nowadays, empty `/useskill` or `/useskill [stop|skill|0]` is what makes the command stop the skill.

I am a `/useskill` (empty) user and the recent changes would introduce a regression where the warning is now printed, so we also update the `argc < 2` behavior to not print it.

### DIFF
--- a/GWToolboxdll/Modules/ChatCommands.cpp
+++ b/GWToolboxdll/Modules/ChatCommands.cpp
@@ -1171,25 +1171,27 @@ namespace {
     } skill_to_use;
     const char* useskill_syntax = "'/useskill [skill]' starts using the skill on recharge.\n"
                                   "Use the skill number instead of [skill] (e.g. '/useskill 5').\n"
-                                  "Use empty '/useskill' or '/useskill stop' to stop all.\n"
-                                  "Use '/useskill [skill]' to stop the skill.";
+                                  "Use empty '/useskill' or '/useskill [stop|skill|0]' to stop the skill.";
     void CHAT_CMD_FUNC(CmdUseSkill)
     {
-        if (!IsMapReady() || argc < 2) {
-            Log::Warning(useskill_syntax);
+        if (!IsMapReady()) {
+            return;
+        }
+        if (argc < 2) {
+            skill_to_use.slot = 0;
             return;
         }
         const std::wstring arg1 = TextUtils::ToLower(argv[1]);
         if (arg1 == L"stop" || arg1 == L"off") {
+            skill_to_use.slot = 0;
             return;
         }
         uint32_t num = 0;
         if (!TextUtils::ParseUInt(argv[1], &num) || num > 8) {
             Log::Warning(useskill_syntax);
             return;
-        }
-        if (skill_to_use.slot == num) num = 0;
-        skill_to_use.slot = num;
+        } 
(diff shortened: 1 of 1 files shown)


GWToolboxdll/Modules/ChatCommands.cpp
```diff
@@ -1171,11 +1171,13 @@ namespace {
...
     const char* useskill_syntax = "'/useskill [skill]' starts using the skill on recharge.\n"
                                   "Use the skill number instead of [skill] (e.g. '/useskill 5').\n"
-                                  "Use empty '/useskill' or '/useskill stop' to stop all.\n"
-                                  "Use '/useskill [skill]' to stop the skill.";
+                                  "Use empty '/useskill' or '/useskill [stop|skill|0]' to stop the skill.";
     void CHAT_CMD_FUNC(CmdUseSkill)
     {
-        if (!IsMapReady() || argc < 2) {
+        if (argc < 2) {
+            return;
```
henderkes, line 1178: now we return out instead of stopping useskill?

GWToolboxdll/Modules/ChatCommands.cpp
```diff
@@ -1171,11 +1171,13 @@ namespace {
...
+        }
+        if (!IsMapReady()) {
```
henderkes: this makes no sense, why print the usage message when we don't have a map loaded?

gam415: I don't know, I am just preserving the old behavior. If the current behavior is incorrect, why blaming?

GWToolboxdll/Modules/ChatCommands.cpp
gam415: This is the same branch that we had before just without the warning, and the same behavior as what arg1 == L"stop" || arg1 == L"off" does. But yeah actually I tried it, and it doesn't stop the skill and neither does /useskill stop|off. I can update these branches with `skill_to_use.slot = 0`, but that is another fix, for the record.

GWToolboxdll/Modules/ChatCommands.cpp
gam415: I could remove the warning if you want now that the argc check is separate if it doesn't make sense to keep it.

GWToolboxdll/Modules/ChatCommands.cpp
henderkes: The current branch does not do what it promises then. Now it's just a bug where it's suggested to the user that it's a valid command, but it's really just ignored. Please fix it properly.

gam415: It was a bug before my changes in the first place. The branch is updated.

henderkes:
thanks

3vcloud:
I wanted /useskill without any args to output the syntax error - if a command exists, but you're using it wrong or you want to know what it does, this is a good shortcut to find out. I know previously /useskill was a valid command but it kills that shortcut and means users have to trawl through the ui to find the syntax

gam415:
@3vcloud No problem, if not done by the time I get back to the computer I will add the warning back.

PR #1823, opened 2026-04-07
Make VQ toggle button on mission map configurable

gunnarbeutner:
Add a "Show VQ toggle button on mission map" setting, enabled by default.


3vcloud:
If the vanquish tracker has any more features it ought to be its own widget, or maybe rename this one

PR #1824, opened 2026-04-08
Fix crash in DrawEnemyCountLabel when tracked_enemies_by_agent_id is empty

gunnarbeutner:
highest_trackable_agent_id was size_t initialized to 0, so the loop `i <= 0` would execute once even when the vector was empty. Change to int initialized to -1 so all iteration loops naturally skip when no enemies have been tracked yet.


3vcloud:
This is the third amendment to this, I don't think this is the way. I'll look at it tomorrow

PR #1828, opened 2026-04-10
Extract VQ overlay from MissionMapWidget into VanquishOverlayModule

gunnarbeutner:
MissionMapWidget now handles only mission map infrastructure (frame hooking, D3D rendering pipeline, coordinate transforms, minimap lines). All vanquish overlay logic (enemy tracking, fog/exploration, frontier borders, navigation, settings) moves to VanquishOverlayModule.

Notes:

This changes where settings are stored, so anyone who has previously changed any of the VQ overlay settings will have their settings restored to defaults.

### DIFF
--- a/GWToolboxdll/Widgets/MissionMapWidget.cpp
+++ b/GWToolboxdll/Widgets/MissionMapWidget.cpp
@@ -22,6 +22,8 @@ namespace {
     bool draw_all_terrain_lines = false;
     bool draw_all_minimap_lines = true;
 
+    std::vector<MissionMapWidget::DrawCallback> draw_callbacks;
+
     // Pixel-to-game-unit scale — converts pixel thickness to game units
     float cached_px_to_game = 1.f;
 
@@ -243,6 +245,10 @@ namespace {
         dx_device->SetTransform(D3DTS_PROJECTION, &ortho);
 
         minimap_lines.Render(dx_device);
+
+        for (const auto& cb : draw_callbacks) {
+            cb(dx_device);
+        }
     }
 
     void EnqueueMinimapLines(GW::Constants::MapID map_id)
@@ -347,3 +353,6 @@ GW::Vec2f MissionMapWidget::GetBottomRight() { return mission_map_bottom_right; 
 GW::UI::Frame* MissionMapWidget::GetFrame() { return mission_map_frame; }
 float MissionMapWidget::GetPxToGame() { return cached_px_to_game; }
 float MissionMapWidget::GetZoom() { return mission_map_zoom; }
+
+void MissionMapWidget::AddDrawCallback(DrawCallback cb) { draw_callbacks.push_back(cb); }
+void MissionMapWidget::RemoveDrawCallback(DrawCallback cb) { std::erase(draw_callbacks, cb); }
--- a/GWToolboxdll/Widgets/MissionMapWidget.cpp
+++ b/GWToolboxdll/Widgets/MissionMapWidget.cpp
@@ -10,7 +10,6 @@
 
 #include <ImGuiAddons.h>
 #include <Modules/QuestModule.h>
-#include <Widgets/VanquishMapOverlayWidget.h>
 #include <Widgets/Minimap/Minimap.h>
 #include <Widgets/MissionMapWidget.h>
 #include <Widgets/WorldMapWidget.h>
@@ -23,6 +22,7 @@ namespace {
     bool draw_all_minimap_lines = true;
 
     std::vector<MissionMapWidget::DrawCallback> draw_callbacks;
+    std::vector<MissionMapWidget::ContextMenuCallback> context_menu_callbacks;
 
     // Pixel-to-game-unit scale — converts pixel thickness to game units
     float cached_px_to_game = 1.f;
@@ -199,19 +199,20 @@ namespace {
         }
 #endif
         if (ImGui::Button("Place Marker")) {
-            VanquishMapOverlayWidget::StopNavigating();
             GW::GameThread::Enqueue([] {
                 QuestModule::SetCustomQuestMarker(world_map_click_pos, true);
             });
             return false;
         }
-        if (QuestModule::GetCustomQuestMarker() && !VanquishMapOverlayWidget::IsNavigating()) {
+        if (QuestModule::GetCustomQuestMarker()) {
             if (ImGui::Button("Remove Marker")) {
                 GW::GameThread::Enqueue([] { QuestModule::ClearCustomQuestMarker(); });
                 return false;
             }
         }
-        if (!VanquishMapOverlayWidget::ContextMenuItems()) return false;
+        for (const auto& cb : context_menu_callbacks) {
+            if (!cb()) return false;
+        }
 
         return true;
     }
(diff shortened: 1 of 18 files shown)


3vcloud:
This should be a widget, then you'd have your own Draw() call to do your rendering in much the same way as the mission map widget; detach them both but expose the mission map related info you need via a public static funciton on mission map widget.

Anything that isn't a widget or a window should be able to work without drawing anything; if you need access to the device to do your stuff, you're either a widget or a window

gunnarbeutner:
Is there anything else you'd want me to change for this one?

3vcloud:
Sorry, been busy.

Use getagentmatchesflags for the dead check - this is what the game does.

I'll check the rest of it in a day or so

3vcloud:
If only 2 values in the ortho projection ever change, we should amend the logic to just update those 2 values inside a static ortho struct in private namespace to avoid the copy. I don't know if it would optimise it out in release, but we don't need to call a function that creates a new ortho object every time

GWToolboxdll/Widgets/VanquishMapOverlayWidget.cpp
```diff
@@ -490,7 +490,7 @@ namespace {
             }
             const auto* living = agent->GetAsAgentLiving();
 
-            if (!living || living->allegiance != GW::Constants::Allegiance::Enemy || !living->GetIsAlive()) {
+            if (!living || !GW::Agents::GetAgentMatchesFlags(agent, GW::TargetFilter::Enemies)) {
```
3vcloud: With GetAgentMatchesFlags you don't need to check if the agent is living or null either - this check is done implicitly inside the gw code. A single `!GW::Agents::GetAgentMatchesFlags(agent, GW::TargetFilter::Enemies)` at line 491 would be enough

3vcloud: The NPC check below this can also be wrapped up in the same function call; check the GW::TargetFilter constexpr's to see how to exclude spirits and minions

GWToolboxdll/Widgets/VanquishMapOverlayWidget.cpp
```diff
@@ -0,0 +1,841 @@
...
+            tracked_enemies_instance_type = instance_type;
+        }
+
+        const auto* player = GW::Agents::GetControlledCharacter();
+        const auto player_pos = GW::PlayerMgr::GetPlayerPosition();
+        const auto* agents = player_pos ? GW::Agents::GetAgentArray() : nullptr;
+        if (!agents) return;
+
+        static bool was_dead = false;
+        const auto* player_living = player ? player->GetAsAgentLiving() : nullptr;
```
3vcloud: Here should also use GW::Agents::GetAgentMatchesFlags using the player include flag and the exclude dead player flag

GWToolboxdll/Widgets/VanquishMapOverlayWidget.cpp
```diff
@@ -0,0 +1,841 @@
...
+        const auto instance_type = GW::Map::GetInstanceType();
+        if (map_id != tracked_enemies_map_id || instance_type != tracked_enemies_instance_type) {
+            tracked_enemies_by_agent_id.assign(tracked_enemies_by_agent_id.size(), {});
+            tracked_enemies_map_id = map_id;
+            highest_trackable_agent_id = 0;
+            tracked_enemies_instance_type = instance_type;
+        }
+
+        const auto* player = GW::Agents::GetControlledCharacter();
+        const auto player_pos = GW::PlayerMgr::GetPlayerPosition();
```
3vcloud, line 473: With the below changes, I don't thin we need player anymore, but we do still need player_pos - GW::PlayerMgr::GetPlayerPosition I think is a shorthand to avoid needing the nullptr check

GWToolboxdll/Widgets/VanquishMapOverlayWidget.cpp
```diff
@@ -0,0 +1,841 @@
...
+void VanquishMapOverlayWidget::Draw(IDirect3DDevice9* dx_device)
+{
+    if (!visible || !MissionMapWidget::IsRenderReady() || !ToolboxUtils::IsExplorable()) return;
+
+    const auto& gameToScreen = MissionMapWidget::GetGameToScreenMatrix();
+    const auto mm_top_left = MissionMapWidget::GetTopLeft();
+    const auto mm_bottom_right = MissionMapWidget::GetBottomRight();
+
+    if (should_draw) {
+        const D3DStateGuard guard(dx_device);
```
3vcloud: We're doing this twice now; once for the minimap lines in the MissionMapWidget, and again here. Instead, consider exposing something like EnqueueMissionMapDraw() on the MissionMapWidget, which would already have these render states, transforms and scissor set up and would mean that all this function would need to do is call the Render().

I would potentially mean that later in life we could easily re-render this widget again if we wanted to at a different scale/position/crop without too much fuss.

GWToolboxdll/Widgets/MissionMapWidget.cpp
```diff
@@ -820,115 +197,23 @@ class D3DFogBuffer : public D3DTriangleBuffer {
         }
 #endif
         if (ImGui::Button("Place Marker")) {
-            nav_active = false;
-            nav_target_pos = {};
-            nav_marker_pos = {};
+            VanquishMapOverlayWidget::StopNavigating();
             GW::GameThread::Enqueue([] {
```
3vcloud, line 202: Instead of having to explicitly mention the vanquish map widget in here, consider adding a hook in the QuestModule called OnCustomMarkerChanged, or maybe even see if you can hook into the UIMessage broadcast from the game when the quest marker is naturally updated; compare the quest id and coords against what we expect the vanquish overlay to be - if its different, we know something else has hijacked the custom marker so we're not navigating anymore.

We want to decouple the modules in a way that we can disable one without impacting the other

GWToolboxdll/Widgets/VanquishWidget.cpp
```diff
@@ -57,7 +57,7 @@ void VanquishWidget::Draw(IDirect3DDevice9*)
         const ImVec2 max(min.x + size.x, min.y + size.y);
         if (ctrl_pressed && ImGui::IsMouseReleased(0) && ImGui::IsMouseHoveringRect(min, max)) {
             int alive = 0, stale = 0;
-            MissionMapWidget::GetTrackedEnemyCounts(alive, stale);
+            VanquishMapOverlayWidget::GetTrackedEnemyCounts(alive, stale);
```
3vcloud, line 60: Should VanquishWidget and VanquishMapOverlayWidget be combined?

gunnarbeutner:
Cool, I'll probably have time tomorrow to review your suggestions and update the PR. Thank you!

GWToolboxdll/Widgets/VanquishWidget.cpp
gunnarbeutner: I'd probably keep them separate. I guess not everyone would necessarily want to have both active at the same time.

gunnarbeutner:
Oops, I forgot to remove my (local) VS18/hot-reload changes.

gunnarbeutner:
So apparently with the changes to how enemies are detected it's now also counting spirits. I'll fix that in a bit.

PR #1829, opened 2026-04-10
Add hot reload support and deploy script

gunnarbeutner:
Launcher: /hotreload <dll-path> signals running toolbox instances to unload via a named event, waits for unload, copies the new DLL, and re-injects into all GW processes.

DLL: creates a manual-reset event "GWToolboxHotReload" on init and polls it each frame to trigger graceful unloading.

Includes deploy.bat convenience script for build + hot reload. Supports config argument: deploy.bat Debug or deploy.bat RelWithDebInfo.

### DIFF
--- a/GWToolbox/Settings.cpp
+++ b/GWToolbox/Settings.cpp
@@ -24,6 +24,7 @@ void PrintUsage(const bool terminate)
             "    /localdll                  Check launcher directory for toolbox dll, won't try to install or update\n\n"
 
             "    /pid <process id>          Process id of the target in which to inject\n"
+            "    /hotreload <dll path>      Signal running toolbox to unload, copy DLL, and re-inject\n"
     );
 
     if (terminate) {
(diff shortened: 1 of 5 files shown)


gunnarbeutner:
Dunno if you want to merge that, but it's something I've found useful locally.

henderkes:
If you really want this the correct way to go is using a hot reload library, if vs' default functionality is not enough. Ejecting & injecting is often times problematic.

gunnarbeutner:
Oh true, it's just that I found VS' native hot-reload stuff to be less than useful - considering how restrictive it is re. some changes. Is it quick and dirty? Sure, and it does occasionally crash. But at least for me it beats having to restart GW after every couple of changes.

3vcloud:
Whats wrong with just closing toolbox, recompiling, then running gwtoolbox.exe again with the /localdll argument?

3vcloud:
image

PR #1830, opened 2026-04-10
Fix Debug crash in LoginModule when -character arg is absent

gunnarbeutner:
GetCommandLinePref returning false is a valid case, not an error. The DEBUG_ASSERT also made the call a no-op in Release builds.

### DIFF
--- a/GWToolboxdll/Modules/LoginModule.cpp
+++ b/GWToolboxdll/Modules/LoginModule.cpp
@@ -25,7 +25,8 @@ namespace {
         GW::Hook::EnterHook();
         wchar_t* parameter_value = GetStringParameter_Ret(param_id);
         wchar_t* cmp = nullptr;
-        DEBUG_ASSERT(GW::UI::GetCommandLinePref(L"character", &cmp));
+        [[maybe_unused]] const bool ok = GW::UI::GetCommandLinePref(L"character", &cmp);
+        DEBUG_ASSERT(ok);
         if (cmp == parameter_value) {
             // charname parameter
             original_charname_parameter = parameter_value;


3vcloud:
It is an error

gunnarbeutner:
Hm OK, for some reason `GW::UI::GetCommandLinePref` returns false here, when starting GW with gwlauncher  (empty character name). Also, `DEBUG_ASSERT(...)` gets optimized out in release builds, so maybe the check should be separate from the function call:

```
#ifdef _DEBUG
#define DEBUG_ASSERT(expr) ((void)(!!(expr) || (Log::FatalAssert(#expr, __FILE__, (unsigned)__LINE__), 0)))
#else
#define DEBUG_ASSERT
#endif
```

henderkes:
Well we wouldn't want to crash users gw in release mode unless necessary. That's why assertions only make debug mode crash.

gunnarbeutner:
Nah, that's not my point. `DEBUG_ASSERT(GW::UI::GetCommandLinePref(L"character", &cmp));` "disappears" completely in release builds, including the call to `GetCommandLinePref`. That's not what's intended here, is it?

henderkes:
Oh yeah that doesn't seem right. We want to assign to a temp var in debug and assert on that though.

gunnarbeutner:
Cool, I'll update the PR. That still leaves me with `GetCommandLinePref` somehow returning false though. Maybe gwlauncher is passing an empty string as an argument to `-character` and `GetCommandLinePref` doesn't like that, I'll have to check.

henderkes:
It shouldn't pass -character at all anymore if there wasn't one entered.

henderkes:
this will fail without [[maybe_unused]]

henderkes (approved):

PR #1833, opened 2026-04-12
update imgui to 1.92-docking, introduces dynamic font scaling

henderkes:
image


3vcloud:
There's a lot of places where you've had to swap out global font scale for font scale main because they're not using the FontScale() function directly - is it worth refactoring those lines, and is there any merit on having the imgui font scale cached for the duration of the frame in this function?

henderkes:
We can actually just go to `FontScale()` as that does nothing but return the styles font scale. Good catch.

> is there any merit on having the imgui font scale cached for the duration of the frame in this function?

No, the function will most likely get inlined in Release builds and is just a pointer deref.

3vcloud:
Ok, and it's also worth checking that fa icons are scaled up too - I don't think they were available before over a certain size

henderkes:
They are, check screenshot above :)

3vcloud:
So does imgui use the same glyphs and manages the different font sizes without our input now? Does I still show icons at 3x?

henderkes:
Yeah everything should still look the same. The change is that it loads the font once, but doesn't rasterize it directly, so it's not bound to a single size. Instead the glyphs are rasterized and cached on demand I think. It's basically the same as if we loaded our fonts at every possible font size, always perfectly crisp.

You can now just set the font size and use font scale to get a larger font for something instead of switching the font for that.

As a nice side effect, our font loading is also much faster now. There's probably an option to prebake too if we really needed to, but I don't see a reason. Slowdown would be for a single frame a font gyph is displayed in a size.

3vcloud:
Thanks for putting in the time

PR #1841, opened 2026-04-18
fix: ensure FillCircle are drawn on Minimap

gam415:
Prior, FillCircle markers were properly rendered on terrain (if enabled), but would not show up on Minimap anymore, even with an alpha component. I did not investigate when this started happening.

Since all the mutation methods check for `vertices.empty()`, we need to initialize `D3DFillCircle` with the non-default constructor. With this, FillCircle markers properly render on the Minimap.

### DIFF
--- a/GWToolboxdll/Widgets/Minimap/CustomRenderer.h
+++ b/GWToolboxdll/Widgets/Minimap/CustomRenderer.h
@@ -40,7 +40,7 @@ class CustomRenderer : public D3DVertexBuffer {
     private:
         void SyncGeometry();
 
-        D3DFillCircle fill_circle;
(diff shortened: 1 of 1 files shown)


3vcloud:
Seems odd, I tested this with eoe and it was fine - are the spirit agents from the agent renderer not d3dfillcircle?

gam415:
I actually did not check before sending this because through playing I know that spirits ranges aren't broken. I just checked, and it seems spirit range circles are rendered with a `Shape_t` in `AgentRenderer` with vertices immediately added to them, they do not seem to be `D3DFillCircle` like `CustomRenderer` uses.

PR #1842, opened 2026-04-19
feat: "prefer same pcons types" on refill

gam415:
With the current refill algorithm, when we have several of the same pcons' type in storage, the first stack is always selected for refill. But this causes characters' inventory to lose space as it will frequently pull in another pcon type instead of the pcon type already in inventory, for example:

pcons

With Lunars (Horse) and Lunars (Rabbit), if I have Lunars (Rabbit) in my inventory but the first stack in storage in Lunars (Horse), I end up losing inventory space with the auto-refill (100 Rabbit, 50 Horse).

One solution is to constantly monitor storage and move pcons around to be properly prioritized by the refiller, but that's time costly with multiple pcons to constantly sort (Alcohol, Exp scrolls, City pcons, Lunars).

With this new option enabled, we first scan the player's inventory for pcons of the given subclass (`PointsPerUse >= 1`), and collect a priority list of distinct model_ids in the same order for that pcon subclass (Lunars/Alc/Scroll...). Then when we refill the bags, if we have a priority list, we search the storage taking that list in consideration.

In the above example, this results in my inventory containing 150 Rabbit Lunars, thus only keeping a single stack in inventory.

### DIFF
--- a/GWToolboxdll/Windows/Pcons.cpp
+++ b/GWToolboxdll/Windows/Pcons.cpp
@@ -400,6 +400,36 @@ void Pcon::Refill(const bool do_refill)
     ResetCounts();
 }
 
+std::vector<DWORD> Pcon::GetPrioritizedModelIdsFromInventory() const
+{
+    std::vector<DWORD> prioritized_ids;
+    GW::Bag** bags = GW::Items::GetBagArray();
+    if (!bags) {
+        return prioritized_ids;
+    }
+    for (auto bagIndex = static_cast<size_t>(GW::Constants::Bag::Backpack);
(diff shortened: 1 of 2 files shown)


3vcloud:
Is there any situation where we would need the setting?

gam415:
It's mostly for preserving the previous behavior by default, but since it's not strictly a breaking change I think we could do away with it and update the default algorithm too if you prefer.

3vcloud:
Yeh I'm thinking unless there's a good use case, replace the logic to reduce code debt

gam415:
Ok, I've updated the commit and it now replaces the old refill algorithm.

PR #1849, opened 2026-04-24
feat: refill pcons when entering outpost and enabled

gam415:
I have heard from several people now that this is the behavior they are expecting out of this feature, but they think it is broken or buggy.

* Expected behavior: Entering outpost with pcons enabled -> refill triggers
* Current behavior: Entering outpost with pcons enabled -> disable pcons -> enable pcons -> refill triggers

This implemented what these users are expecting. Personally it's never really bothered me, but I can see their point.

### DIFF
--- a/GWToolboxdll/Windows/Pcons.cpp
+++ b/GWToolboxdll/Windows/Pcons.cpp
@@ -187,7 +187,10 @@ void Pcon::Update(int delay)
         maptype = GW::Map::GetInstanceType();
         SetPlayerName();
         ResetCounts();
-        Refill(refill_if_below_threshold && IsEnabled() && PconsWindow::Instance().GetEnabled());
+        Refill(false);
+        if (maptype == GW::Constants::InstanceType::Outpost) {
+            Refill(refill_if_below_threshold && IsEnabled() && PconsWindow::Instance().GetEnabled());
+        }
     }
     // Refill pcons if needed.
     UpdateRefill();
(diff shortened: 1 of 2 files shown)


henderkes:
When did this get lost? Are you sure it's not somewhere in the code and not just broken?

gam415:
Did it get lost? I have no idea how it used to behave in the past, but this is how it behaves now and it has been like this for a while now. Only disabling/re-enabling pcons will trigger the refill.

henderkes:
> Are you sure it's not somewhere in the code and not just broken?

Hence why I'm asking this.

gam415:
Yeah, pretty sure.

henderkes:
Well I guess I'll have to look then...

gam415:
It seems maybe https://github.com/gwdevhub/GWToolboxpp/commit/55593c17 broke it back in 2021.

gam415:
based on the approach that was in https://github.com/gwdevhub/GWToolboxpp/commit/55593c17, I pushed a second commit which is a fix rather than a feat, and should mirror the old approach. PR can be squashed if the second commit is preferred.

haxgw:
No idea if it helps but this broke very recently, I can't say exactly when but it has worked in the last couple of months as far as I know.

gam415:
Really? I don't even recall it working in the one year that I have been using Toolbox - but maybe I am wrong.

3vcloud:
Only just managed to sit down and review the pcons code that was merged in recently, along with this issue. The refill process is correctly triggering and gets past this noop process; theres more going on than this, and this fix would simply bomb out of any pending refills, messing up any reserved refill slots that anything else is using - theres a reason this was so complicated in the first place.

Theres also some weird logic around the priority model ids thing that was recently merged in, including at least 1 memory leak, so I need to revisit the pcons code.

PR #1853, opened 2026-04-26
Make VQ overlay frontier border update incremental

gunnarbeutner:
Instead of scanning the entire walkable grid when cells are explored, only recompute frontier edges for newly explored cells and their immediate neighbors. Store edges per-cell in a map so incremental updates can modify individual cells without a full rebuild.

Before:

image

After:

image


3vcloud:
Love that module, can we get it in?

PR #1854, opened 2026-04-26
Add performance monitor window

gunnarbeutner:
Per-module Update/Draw timing with QPC, Present hook for GPU time, 5-second rolling window stats. Profiling only active while the window is visible (no overhead otherwise).


gunnarbeutner:
I need to test this first before this can be merged. My local branch has a slightly different `Draw` function and I just rebased this against dev.

PR #1858, opened 2026-04-27
Fix UB and weak hash combining in ItemHash and SlotHash

gunnarbeutner:
The old code only hashed Data1/Data2/Data3, missing the 8-byte Data4 field entirely. It also used `reinterpret_cast<uint64_t*>` on the GUID (strict aliasing violation + potential misalignment) and XOR-with-shift for combining. Replace with `std::bit_cast` on individual GUID fields and a proper golden-ratio combiner (hash_combine).

### DIFF
--- a/GWToolboxdll/Windows/AccountInventoryWindow.cpp
+++ b/GWToolboxdll/Windows/AccountInventoryWindow.cpp
@@ -1,4 +1,5 @@
 #include "stdafx.h"
+#include <bit>
 #include <fileapi.h>
 
 #include <GWCA/GameEntities/Agent.h>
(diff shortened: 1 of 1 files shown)


3vcloud:
The first 3 parts of the guid make 64 bits, and it wasn't worth bothering with the last 8 bits. Not sure how it could be misaligned when casting from uuid* to uint64*. Guid is 32 + 16 + 16 + 8

gunnarbeutner:
```
typedef struct _GUID {
    unsigned long  Data1;
    unsigned short Data2;
    unsigned short Data3;
    unsigned char  Data4[ 8 ];
} GUID;
```

The natural alignment for that struct is 4 (due to the `unsigned long` member), while the natural alignment for `uint64_t` is 8. Which means a `UUID` object might not necessarily be aligned on an 8-byte boundary. You can get away with unaligned reads on x86, but it's actually undefined behavior.

This also currently violates the strict aliasing rule because the types (`UUID` and `uint64_t`) are unrelated.

3vcloud:
I'd agree if we were assigning an address, but we're not - we don't care about alignment, we just want the compiler to stop moaning and do its job by reading the value from a specific address as a specific type that its been told it can handle for the hashing process. Doesn't matter if I told the hasher it was a uint8, uint16, uint32 of uint64; the starting address is the same, the types are basic, and the difference is the detail we want to get into when hashing a uuid alongside all of the other painful stuff being put into this hash key like character name etc.

We need to revisit the process of comparing item a to item b anyway, but this module is so much code that its not worth the headache atm

PR #1868, opened 2026-04-29
Fix CMake preset to require VS 2022 instead of VS 2026

gunnarbeutner:


henderkes:
I just updated to 2026 and uninstalled 22, can we leave it there now?

gunnarbeutner:
I don't mind either way, want me to update the README instead? Maybe even mention how to use `-G`?

henderkes:
The preset does it so shouldn't need any readme change? Do we mention 17 somewhere?

gunnarbeutner:
Yeah, it suggests downloading VS 2022. I've updated the PR to just update the README.

gunnarbeutner:
I've also updated the GitHub workflow to use VS2026.

henderkes:
Almost hood, but let's not explain how to use VS2022. People will have to update when we no longer test vs 22.

henderkes (approved):

PR #1873, opened 2026-04-29
Fix VQ overlay transform using wrong position source in spectator mode

gunnarbeutner:


3vcloud:
Probably needs disabling unless you're alive - would get too messy with exploration etc

gunnarbeutner:
That's kinda what we're doing atm anyway, at least re. tracking enemies/exploration. I.e. when you're dead we're not updating tracked enemies / explored area (and I'm not even sure if that's the correct behavior). However, when spectating someone else the map overlay wouldn't correctly move with the base map layer - i.e. the positions for the overlay and the base map would get out of sync:

Spectating myself (correct):

gw704

Spectating someone else (incorrect transform):

gw705

PR #1880, opened 2026-05-01
Filter VQ overlay to reachable areas using blockedPlanes

gunnarbeutner:
BFS from the player's trapezoid through adjacency and unblocked portals, matching the game's native A* blocking checks:
  - `portal.flags & 0x04` (portal blocked)
  - `blockedPlanes[neighbor_plane] & 1` (plane blocked at runtime)

Monitors `blockedPlanes` for changes and rebuilds when they update.

Known limitation: blockedPlanes are sent by the server as the player approaches gates/walls, not at map load. The overlay starts showing all areas and shrinks as blocking data arrives.

https://github.com/user-attachments/assets/3b4e7195-ffa2-48e7-ab53-c8381b5963af

### DIFF
--- a/GWToolboxdll/Widgets/VanquishMapOverlayWidget.cpp
+++ b/GWToolboxdll/Widgets/VanquishMapOverlayWidget.cpp
@@ -1,6 +1,7 @@
 #include "stdafx.h"
 
 #include <GWCA/Context/GameplayContext.h>
+#include <GWCA/Context/MapContext.h>
 #include <GWCA/GameEntities/Agent.h>
 #include <GWCA/GameEntities/NPC.h>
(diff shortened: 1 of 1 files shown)


gunnarbeutner:
TODO: Maybe render temporarily-inaccessible areas in a different color.

gunnarbeutner:
This is ready to go as far as I'm concerned:

image

henderkes:
I legitimately can't tell what's what. The red areas are inaccessible?

3vcloud:
Be careful not to put too much stuff into a widget thats meant for vanquishing - this would almost be a candidate for a new renderer type that can be transposed onto mission map/compass/world map

PR #1912, opened 2026-05-03
More EncString clean-up

gunnarbeutner:
**Make EncString::wstring/string const and nodiscard**

Lazy-eval members (`decoded_ws`, `decoded_s`, `decoding`, `decoded`, `sanitised`, `pending_ctx_`) are now mutable so `wstring()`/`string()` can be `const`. Both are marked `[[nodiscard]]`.

Add `StartDecode()` for fire-and-forget decode triggering, replacing ~18 callsites that called `wstring()`/`string()` and discarded the
result. Fix const-ref mismatches in Hotkeys.h, EnemyWindow.cpp, and GWMarketWindow.cpp.

**Simplify EncString state: remove decoding and sanitised flags**

Replace the `decoding` flag with `pending_ctx_` nullity check. Sanitise eagerly in `OnStringDecoded` instead of lazily in `wstring()`, removing the `sanitised` flag. Store `should_sanitise` as a configuration bool set by `reset()`. Only decoded remains as mutable state.

**Remove EncString::reset, language, and decode; add move semantics**

Remove `reset()` in favour of move-assigning a fresh `EncString`. Remove `language()` - the language is now a constructor parameter.
Fold `decode()` into `StartDecode()` as the single entry point. Add move constructor/assignment that transfer any in-flight decode.
Store a copy of the encoded string in `DecodeContext` so GW's async decode buffer outlives the `EncString`.

### DIFF
--- a/GWToolboxdll/Modules/ChatCommands.cpp
+++ b/GWToolboxdll/Modules/ChatCommands.cpp
@@ -284,7 +284,7 @@ namespace {
         {
             const auto title_info = GW::PlayerMgr::GetTitleData(title);
             if (title_info)
-                name.reset(title_info->name_id);
+                name = GuiUtils::EncString(title_info->name_id);
         };
         GW::Constants::TitleID title;
         GuiUtils::EncString name;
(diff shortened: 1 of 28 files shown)


3vcloud:
reset() had logic to avoid re-decoding if the encoded value is the same; it looks like this code will now keep asking gw to decode every frame that reset() would have been called - an easy fix, but despite that its now more complicated to follow the logic of what encstring is actually doing in cpp land, with more code debt added to the file.

The previous approach was easier

PR #1915, opened 2026-05-03
feat: expose MinimapRenderer through plugin_base

gam415:
I have a number of out-of-tree plugins and a couple of them draw on the Minimap, currently by replicating the d3d state that GWToolbox sets up and cleaning up after themselves which is cumbersome and doesn't detect some user-applied transformations like minimap rotation state, etc. I wanted to switch to the new `MinimapRenderer` approach from [17191bcb](https://github.com/gwdevhub/GWToolboxpp/commit/17191bcb) but quickly found out that it would not work unless I pull my plugins in-tree and/or build all of GWToolboxdll which would nuke my build time. I know there are other out-of-tree plugins out there as well that could benefit from this.

**Cause:**

`Minimap.h` depends on `gwtoolboxdll_export.h` which is cmake-generated and only exists in the build directory. Out-of-tree plugins couldn't use `MinimapRenderer` without building GWToolboxdll first (which pulls many dependencies and SDKs, significantly slowing down build time for out-of-tree plugins).

The goal is to not have to pull `Minimap.h` in plugins extending the Minimap anymore.

**Changes:**

- Add `GWToolboxdll/RectF.h`: RectF extracted from `GuiUtils.h` into a standalone header with no heavy dependencies
- Add `GWToolboxdll/MinimapPlugin.h`: `RectF`, `MinimapRenderContext` and `MinimapRenderer` with a self-contained export macro; `RegisterRenderer/UnregisterRenderer` are dllexported statics on `MinimapRenderer`, defined in `Minimap.cpp`. Minimap::RegisterRenderer/UnregisterRenderer are now private, accessible only via `MinimapRenderer`.
- `plugin_base`: expose `RectF.h` and `MinimapPlugin.h` via target_sources
- `anchor_point` and `Translation` are zero-initialized by default

**Result:**

Plugins can import `MinimapPlugin.h` (from `plugin_base` which doesn't require GWToolboxdll to be built) and go:
```c++
#include <MinimapPlugin.h> // ships in source tree, no build-tree dependency

class MyRenderer : public MinimapRenderer {
public:
    void RenderMinimap(IDirect3DDevice9* device, const MinimapRenderContext& ctx) override {
        // draw stuff
    }
};

class MyPlugin : public ToolboxPlugin {
    MyRenderer renderer;
    void Initialize(...) override {
        MinimapRenderer::RegisterRenderer(&renderer);
    }
    void SignalTerminate() override {
        MinimapRenderer::UnregisterRenderer(&renderer);
    }
};
```

### DIFF
--- a/GWToolboxdll/MinimapPlugin.h
+++ b/GWToolboxdll/MinimapPlugin.h
@@ -1,0 +1,64 @@
+#pragma once
+
+#include <Windows.h>
+#include <d3d9.h>
+#include <GWCA/GameContainers/GamePos.h>
+#include <RectF.h>
+
+#ifndef GWTOOLBOXDLL_MINIMAP_EXPORT
+#  ifdef GWToolboxdll_EXPORTS
+#    define GWTOOLBOXDLL_MINIMAP_EXPORT __declspec(dllexport)
+#  else
+#    define GWTOOLBOXDLL_MINIMAP_EXPORT __declspec(dllimport)
+#  endif
+#endif
+
+// Context structure that encapsulates all rendering parameters
+struct MinimapRenderContext : RectF {
+    // Position and size
+    ImVec2 anchor_point = {}; // Center of minimap - this may not necessarily be the center point of the clipping rect
+
+    // Camera/view parameters
+    GW::Vec2f translation = {}; // World-space translation (for panning)
+    float zoom_scale = 1.f; // Zoom level (scale factor)
+    float rotation = 1.5708f; // Map rotation in radians
+
+    float base_scale = 1.f; // The size (in px) of the base scale to use before zooming. Minimap sets this to size.x
+
+    // Visual options
+    bool circular_map = false; // Whether to render as circle or square
+    bool draw_center_marker = false; // Whether to draw center marker when panned
+    D3DCOLOR background_color = D3DCOLOR_ARGB(50, 0, 0, 0); // Background color (or 0 to use renderer's default)
+    D3DCOLOR foreground_color = D3DCOLOR_ARGB(0xff, 0xe0, 0xe0, 0xe0); // Foreground color (or 0 to use renderer's default)
+    D3DCOLOR shadow_color = 0; // Drop shadow for foreground color
+    D3DCOLOR cardinal_color = 0;
+
+    bool draw_ranges = false;
+
+    RECT rect() const
+    {
+        return {static_cast<LONG>(top_left.x), static_cast<LONG>(top_left.y), static_cast<LONG>(bottom_right.x), static_cast<LONG>(bottom_right.y)};
+    }
+};
+
+// subclass this structure and register an instance to have plugins draw extra stuff on a minimap
+// Vtable layout (do not change): slot 0 = ~MinimapRenderer, slot 1 = RenderMinimap.
+class MinimapRenderer {
+public:
+    MinimapRenderer() = default;
+    virtual ~MinimapRenderer() = default;
+
+    // minimap holds pointers to registered renderers, so move is forbidden
+    MinimapRenderer(const MinimapRenderer&) = delete;
+    MinimapRenderer(MinimapRenderer&&) = delete;
+    MinimapRenderer& operator=(const MinimapRenderer&) = delete;
+    MinimapRenderer& operator=(MinimapRenderer&&) = delete;
+
+    // implement this to draw your data on the minimap
+    // the view matrix is set up to transform game world positions to minimap appropriately, make sure to restore it if you touch it
+    // the world matrix is undefined and can be left in any state
+    virtual void RenderMinimap(IDirect3DDevice9* device, const MinimapRenderContext& context) = 0;
+
+    GWTOOLBOXDLL_MINIMAP_EXPORT static void RegisterRenderer(MinimapRenderer* renderer);
+    GWTOOLBOXDLL_MINIMAP_EXPORT static void UnregisterRenderer(MinimapRenderer* renderer);
+};
--- a/GWToolboxdll/RectF.h
+++ b/GWToolboxdll/RectF.h
@@ -1,0 +1,58 @@
+#pragma once
+
+#include <Windows.h>
+#include <imgui.h>
+
+struct RectF {
+    ImVec2 top_left;
+    ImVec2 bottom_right;
+
+    constexpr RectF(const RECT& rect)
+        : top_left({static_cast<float>(rect.left), static_cast<float>(rect.top)}),
+          bottom_right({static_cast<float>(rect.right), static_cast<float>(rect.bottom)}) {}
+
+    constexpr RectF(const ImVec2& _top_left, const ImVec2& _bottom_right)
+        : top_left(_top_left), bottom_right(_bottom_right) {}
+
+    RectF() = default;
+
+    ImVec2 size() const
+    {
+        return {width(), height()};
+    }
+
+    float width() const
+    {
+        return bottom_right.x - top_left.x;
+    }
+
+    float height() const
+    {
+        return bottom_right.y - top_left.y;
+    }
+
+    void move_to(const ImVec2& new_top_left)
+    {
+        const auto diff_x = new_top_left.x - top_left.x;
+        const auto diff_y = new_top_left.y - top_left.y;
+        top_left.x += diff_x;
+        top_left.y += diff_y;
+        bottom_right.x += diff_x;
+        bottom_right.y += diff_y;
+    }
+
+    void resize(const ImVec2& new_size)
+    {
+        const auto old_size = size();
+        const auto diff_x = new_size.x - old_size.x;
+        const auto diff_y = new_size.y - old_size.y;
+        bottom_right.x += diff_x;
+        bottom_right.y += diff_y;
+    }
+
+    bool contains(const ImVec2& point) const
+    {
+        return point.x >= top_left.x && point.x <= bottom_right.x &&
+               point.y >= top_left.y && point.y <= bottom_right.y;
+    }
+};
--- a/GWToolboxdll/Utils/GuiUtils.h
+++ b/GWToolboxdll/Utils/GuiUtils.h
@@ -4,64 +4,11 @@
 #include <ImGuiAddons.h>
 #include <nlohmann/json.hpp>
 #include <ToolboxIni.h>
+#include <RectF.h>
 
 namespace GW::Constants {
     enum class Language;
 }
-
-struct RectF {
-    ImVec2 top_left;
-    ImVec2 bottom_right;
-
-    constexpr RectF(const RECT& rect)
-        : top_left({static_cast<float>(rect.left), static_cast<float>(rect.top)}),
-          bottom_right({static_cast<float>(rect.right), static_cast<float>(rect.bottom)}) {}
-
-    constexpr RectF(const ImVec2& _top_left, const ImVec2& _bottom_right)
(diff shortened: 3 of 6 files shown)


henderkes:
I get why you want this, but that's really not what the plugin system is made for and wouldn't even integrate particularly well with Toolboxes own modules 

The whole TAS crap is a real cancer to the game. Any minimap changes useful to the public are welcome to be added to toolbox directly, so I see no reason to support this.

gam415:
I'm not sure that the "TAS crap" has to do with this tbh, this is just about friendly extensibility, there are lots of non TAS-related use-cases for extending the minimap one could come up with. 

> that's really not what the plugin system is made for and wouldn't even integrate particularly well with Toolboxes own modules

What is the plugin system made for then? And how does this not integrate well with Toolbox's own modules? It's literally backwards compatible and in fact, I don't even see the MinimapRenderer used anywhere in the current Toolbox codebase.

gam415:
The game's population and the Toolbox userbase will drop by half or more the day the "TAS crap" gets removed, by the way... What is an already dying game will just be even more desert.

henderkes:
> I'm not sure that the "TAS crap" has to do with this tbh, this is just about friendly extensibility, there are lots of non TAS-related use-cases for extending the minimap one could come up with.

And 99% of them would be better off being upstreamed into toolbox. 

> What is the plugin system made for then?

Initially, to appease users who were not updating their toolbox but wasting our time with support enquiries. It's most certainly not made to fragment the user- and feature base more.

> The game's population and the Toolbox userbase will drop by half or more the day the "TAS crap" gets removed, by the way... What is an already dying game will just be even more desert.

You overestimate the amount of people who will quit the game over it and you *vastly* underestimate the amount of toolbox users and gw players, especially now. But that's completely irrelevant, because it's just a fucking awful idea to implement useful features in plugins that every user is *strongly discouraged from using and warns them that their account can be banned because using any toolbox plugins is expressly disallowed*.

gam415:
Being part of the SC community who is about half or more of all of the players in Kamadan and Embark Beach, and constantly part of the TAS conversations in these alliances, let's just agree to disagree. I guess we'll never know. Alright, no problem, closing this then.

3vcloud:
I'm getting a bit annoyed at people throwing the baby out with the bathwater in arguments on github before I've even had a chance to make my own reply.

@gam415 's past contributions are pretty much directly from the TAS version, so I appreciate that theres a user base for it and making PR's like this one do help bring the communities together and helps the main toolbox fork too.

I deliberately moved the created MinimapContext to use this approach for pretty much this exact reason - someone was making a plugin and wanted to render onto the TB Minimap without having to reinvent the wheel, so this PR is actually spot on for what I wanted to achieve.

The plugin module was supposed to avoid having an entire fork of toolbox in favour of you just having small dlls that do whatever hacks you want under the agreement that plugins aren't supported. Its a shame that the TAS fork still isn't there yet, and still relies on manual code amends of the toolbox source to do these features rather than hooks/patches in a plugin.

As an aside to this, we keep getting users raising issues/bugs that aren't related to the latest version of toolbox even though the version number matches, because they're opting to use a fork of this repo that distributes builds from a development branch, and rather than waste my time I've just started ignoring or closing issues that I think aren't to do with our version of toolbox, which is a massive shame and could be discounting legitimate bugs with the main repo, but its now the only way I know of to avoid wasting my time debugging something whilst keeping toolbox open source.

henderkes:
> Being part of the SC community 

And you think we aren't? Toolbox was created by speedclearers and is still maintained by us. There are many different communities though, and only very few of them hack the game in ways that arenanet deems as bannable. There are even speedclear players who don't use toolbox at all.

> who is about half or more of all of the players in Kamadan and Embark Beach

Embark maybe, not Kamadan, but we're looking at maybe 5% of toolboxes userbase who use plugins at all. And it would be good if it were fewer.

The point is that everything useful should be *upstreamed*, while everything bannable should preferably not exist at all, because it endangers toolboxes very existence. So I'm not a fan of code changes that are good for nothing but endangering arenanets toleration of toolbox.

gam415:
> @gam415 's past contributions are pretty much directly from the TAS version

Not quite though, I make a distinction between contributions to upstream (here) and the few things that we have in the TAS versions, and I know the difference and never contribute something to TAS if I know it will benefit upstream first (and cascade into TAS). I value both repositories/projects and wouldn't contribute something to TAS if I knew it could belong to upstream instead.

> I deliberately moved the created MinimapContext to use this approach for pretty much this exact reason - someone was making a plugin and wanted to render onto the TB Minimap without having to reinvent the wheel, so this PR is actually spot on for what I wanted to achieve

So, would you be open to this PR? Because the only difference with and without it is to be able to save build time by skipping GWToolboxdll build when extending the Minimap. Otherwise, plugins extending the Minimap will still be achievable, just costing longer builds and GH Actions runs... Hence I didn't really see this feature as anything TAS-related, it really seemed like a genuine upstream improvement to me.

Again, when I come to this repository, I'm not wearing a TAS hat - didn't really appreciate it being brought up in the first place because it mischaracterized the approach I try to follow. Perhaps I took the bait and shouldn't have, apologies for escalating the conflict there. I hope this makes things clearer.

3vcloud:
Not all plugins are hacks, there have been a lot of requests for niche features that I've suggested would be good candidates for being a plugin rather than compiled into the main app.

In this example, if we merged this PR, someone could make their own plugin to basically make a carbon copy of the Minimap on a separate window to the main game. A plugin like that would be a fair amount of code debt, so I'm not sure if it would be a good candidate to be part of the main toolbox code that we'd inherit and have to maintain, but to the right person it could be a real help.

imo this PR isn't TAS related and I think its a good idea to have

gam415:
> So I'm not a fan of code changes that are good for nothing but endangering arenanets toleration of toolbox.

Same argument as in my above message: extending the minimap is already possible (even with or without the `MinimapRenderer` thing), this PR just saves build time for plugins who do, that's all... It seemed like a nicer separation of concerns and a healthy improvement to the codebase.

henderkes:
> In this example, if we merged this PR, someone could make their own plugin to basically make a carbon copy of the Minimap on a separate window to the main game. A plugin like that would be a fair amount of code debt, so I'm not sure if it would be a good candidate to be part of the main toolbox code that we'd inherit and have to maintain, but to the right person it could be a real help.

I think we can all agree that there are legitimate use-cases for plugins that are better not being upstreamed.

> imo this PR isn't TAS related and I think its a good idea to have

My concern is that we shouldn't encourage building things into a plugin when most changes, especially those regarding the minimap, would be better upstreamed. I struggle thinking of use-cases where a plugin should be drawing on the minimap that isn't a good idea to upstream.

> It seemed like a nicer separation of concerns and a healthy improvement to the codebase.

I don't disagree, but refactoring specifically to make plugins easier is not a priority for me.

henderkes:
Perhaps, to illustrate my point a bit better: https://github.com/gam415/GWToolboxPlugins

- EffectsIndicator - should be upstreamed
- LootNotifier - should be upstreamed
- NameObfuscator - good as plugin
- PartyReorder - should be upstreamed
- SafeShadowWalk - good for plugin because it's very niche
- TargetDetector - should be upstreamed
- WeaponRangeIndicator - should be upstreamed

3vcloud:
Yeh I just had a look myself, this is some really awesome work! Its up to you if you want to PR this stuff in or keep maintaining them alongside the main toolbox, but this does emphasize how the plugins module has actually fractured toolbox as a tool to do these things despite what we wanted to do

gam415:
Thank you, and yes I have thought about it, but there are some reasons why they are not, at least at the moment. It could be worth a discussion, more below. If you can believe it, I actually have been in your shoes before having written and maintained one of the big HTTP proxies out there for almost 10 years, and dealing with a plugins system, what users want out of it, our vision for it and what we wanted to have in the main product itself, poorly written plugins blaming the host for X or Y degradation, etc... Although, minus the 3rd party (ANet) and the risk of getting banned for using a plugin in that world, of course... So I understand your goals and concerns. 

As for the plugins and why they aren't upstreamed, as much as I would not be opposed to it, there's also a few reasons why they might not be good candidates that I can list:

1. Most of the time, I didn't even know how the feature would be received upstream - if there is interest, then why not, yes.
2. Having less control over the release cycle can be a hindrance - since the plugins were young (until recently), I had to iterate fast on them and having control over their release cycle was necessary to get some usage and fix whatever bugs would come up relatively fast. I built an update system that notifies users of new features while in-game, and not having to wait for new TB releases is kind of nice for shipping improvements.
3. While the plugins work really well now, I suspect that some would actually be burdensome to bring upstream because they have limitations that I had to circumvent or they would be a lot of code to bring-in, for example:
    - EffectsIndicator: super useful, but can be difficult to configure for most players. They have to use an "Effects Tracker" window to RE the effects triggered by AoE skills, and determining the number of effect packets per skill, AoE size, etc... Not very intuitive.
    - PartyReorder: is very robust and really smart at reordering, my favorite by far. Can detect which reorder to apply based on party composition, has built-in retries for invites, a full-fledged party editor in settings, handles tons and tons of possible edge-cases that can happen during the reorder process. That thing is awesome. But the cost for all that is a little bit over 3500 LOC, which I think could be an argument for not merging this upstream as the cost of maintenance can be really burdensome.
    - WeaponRangeIndicator: is also super nice and very useful for fast tactics and smooth gameplay, but also has some hidden costs under the hood. Namely, there isn't a way to determine the subtype of a bow (as far as I could tell). As a result, bow subtype is determined by a multi-step categorization based on decoded bow names, e.g. "does it have `"flatbow"` in the name?", if not, then "is it one of these other bow names", etc. And to make it work with different languages, I had to hard-code quite a few bow names in there from the wiki list. It works but it's a bit hacky, not sure this would be accepted upstream. I also have a lot of utils libraries for proper async decoding for item names, map names, etc, and a whole lot of other things.

I could see some good candidates, perhaps LootNotifier, although in terms of configuration, users also need to be familiar with `model_file_id` attributes etc, which is perhaps a little too advanced (same kind of UX as EffectsIndicator), maybe not... 

So, for now I wrote them as plugins, and the main reason really being I didn't think they would be accepted upstream. I really think plugins are awesome, personally, but I also understand where you are coming from.

henderkes:
> Having less control over the release cycle can be a hindrance - since the plugins were young (until recently), I had to iterate fast on them and having control over their release cycle was necessary to get some usage and fix whatever bugs would come up relatively fast. I built an update system that notifies users of new features while in-game, and not having to wait for new TB releases is kind of nice for shipping improvements.

I absolutely get the reasoning, but in the process you've trained users who want these additions to install plugins, which in this context may be fine, but with other plugins makes them susceptible to getting their account banned permanently. There's a reason why we removed all the TAS stuff from Toolbox despite some being legitimately useful features that we miss (opening chests from afar, multiple useskill, etc.).

> Most of the time, I didn't even know how the feature would be received upstream - if there is interest, then why not, yes.

We're very liberal with accepting feature additions for that reason. Almost anything that's useful to a number of people is accepted in contribution. We don't want the feature set fragmented in a way that is not beneficial to the long-term health of the project, especially not through a system thats very existence is based on doing forbidden things.

> I actually have been in your shoes before having written and maintained one of the big HTTP proxies out there for almost 10 years

Funnily enough I work with Caddy and maintain FrankenPHP, so I understand where you are coming from. Unfortunately, it's a very different situation with Toolbox and the grey area in which we've operated for over a decade.

As for your other concerns, through the power of Jon, we can probably clean those edges out for bow type and others. For the maintenance burden or users having to figure out ids, we have that in multiple modules. Just having sensible defaults is enough for 99% of users anyway.

PR #1918, opened 2026-05-04
Swap Beknur_Harbor and Beknur_Harbor_outpost enum names

gunnarbeutner:
MapID 457 is an explorable area and 487 is the outpost, but the enum names had them backwards. Verified in-game: standing in the outpost reports MapID 487 with RegionType::Outpost.

Outpost:

gw761

Explorable (during [Assault on Beknur Harbor](https://wiki.guildwars.com/wiki/Assault_on_Beknur_Harbor)):

gw762


henderkes:
will fix in gwca, thanks

PR #1921, opened 2026-05-04
Add per-type size options to agent renderer (Hostile, Neutral, Ally NPC, Quest Giver, Spirit)

### LINKED ISSUE #1572: [Suggestion] Sizable Agent "Ally (NPC Quest Giver)"
Would be very interested to have the "Ally (NPC Quest Giver)" icon to be resizable as well.

3vcloud:
Closes #1572

Adds dedicated size controls for each agent type that has a colour picker in the Agent Colors settings section but previously had no dedicated size: Hostile, Neutral, Ally (player), Ally (NPC), Ally (NPC Quest Giver), Ally (spirit).

The size drag-float appears inline next to each colour picker. All new sizes default to 75 (same as size_default) so existing behaviour is preserved.

Custom agents already had per-agent size support and are unchanged.

### DIFF
--- a/GWToolboxdll/Widgets/Minimap/AgentRenderer.cpp
+++ b/GWToolboxdll/Widgets/Minimap/AgentRenderer.cpp
@@ -442,61 +442,36 @@ void AgentRenderer::DrawSettings()
         {
             const float size_w = ImGui::GetTextLineHeight() * 4.f;
             const float spacing = ImGui::GetStyle().ItemSpacing.x;
-
-            ImGui::SetNextItemWidth(ImGui::GetContentRegionAvail().x - size_w - spacing);
-            Colors::DrawSettingHueWheel("Hostile (>90% HP)", &color_hostile);
-            ImGui::SameLine();
-            ImGui::SetNextItemWidth(size_w);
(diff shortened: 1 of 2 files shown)


GWToolboxdll/Widgets/Minimap/AgentRenderer.cpp
```diff
@@ -421,13 +439,79 @@ void AgentRenderer::DrawSettings()
         Colors::DrawSettingHueWheel("Player (dead)", &color_player_dead);
         Colors::DrawSettingHueWheel("Signpost", &color_signpost);
         Colors::DrawSettingHueWheel("Item", &color_item);
-        Colors::DrawSettingHueWheel("Hostile (>90% HP)", &color_hostile);
+        {
+            const float size_w = ImGui::GetTextLineHeight() * 4.f;
```
3vcloud, line 443: @claude these blocks of code recycle the same values - just get them once

3vcloud: @claude take this further to reduce the code lines - an array of info containing the labels needed and a pointer to the colour and size variables for each type of agent would mean we only need to write the imgui calls once. Consider other places in this module where this could be easily achieved

PR #1895, opened 2026-05-02
Show Nicholas Sandford info in pre-searing item tooltips

henderkes:
## Summary
Hovering a Sandford collectible (Grawl Necklace, Charr Carving, etc.) in pre-searing now adds a tooltip line analogous to Nicholas the Traveler in post-searing:

> "Nicholas Sandford collects 5 of these in N days!"

Sandford's quantity is always 5 per request, so it is hard-coded.

Implementation:
- New `DailyQuests::GetNicholasSandfordItemInfo` returns the matching pre-searing cycle entry by encoded item name.
- New `DailyQuests::GetTimestampFromNicholasSandford` returns the next collection time using the existing pre-searing rotation (24h period, 52-day cycle, epoch 2009-04-09).
- `ItemTooltipModule::AppendNicholasInfo` switches on `GW::Map::IsPreSearing()` so Sandford-only items don't accidentally show a Traveler tooltip in post-searing.

The existing "Show Nicholas the Traveler info in item tooltip" toggle controls both, since it would be confusing to have a separate setting for a feature that's effectively the same thing in pre-searing.

## Test plan
- [ ] In a pre-searing character, hover a Sandford collectible (e.g. Skale Fins) and confirm the new tooltip line appears with the correct day-count
- [ ] Verify the same item in post-searing does NOT show the Sandford line (it shouldn't trigger Traveler either if it's not in his cycle)
- [ ] Confirm the existing Traveler tooltip in post-searing is unchanged

Closes #1749

---
_Generated by [Claude Code](https://claude.ai/code/session_01Wo6qnHxyVrRdgwgY7ZRNnu)_

PR #1924, opened 2026-05-05
Add "show other searing dailies" checkbox to Daily Quests window

3vcloud:
Adds a right-aligned checkbox at the top of the Daily Quests window labeled "Show post searing dailies" (when in pre) or "Show pre searing dailies" (when in post). When checked, a second section appears below the primary daily quests showing the other searing's rotation. Default off, persisted to ini.

Implemented in response to a request on PR #1895.


3vcloud:
@claude with this change, we no longer need show_presearing_dailies_in_window

3vcloud:
@claude this code has added post searing and pre searing dailies as separate tables, which is wrong. We want the data combined with a column per daily type. If we're in pre and we need to show post aswell, then you'd add the post searing columns after the pre searing columns and vice versa.

Also revise the code itself - I feel like theres a lot of duplicated logic to draw these columns/data, and by looping some arrays instead we could reduce lines

PR #1928, opened 2026-05-05
Z-order interleaving between GW and TB windows

gunnarbeutner:
- Click-to-raise works across both GW and Toolbox windows
- GW popups can appear above Toolbox windows and vice versa
- Unpinned Toolbox windows close with Escape (most-recently-focused first)
- Thumbtack pin button in each Toolbox title bar

gw764


henderkes:
This was partially merged, if you still want to make TB windows compositable behind gw ones it should be a simple reimplementation.

PR #1946, opened 2026-05-07
Terminate old gwca.dll before loading toolbox version

### LINKED ISSUE #1945: Terminate old gwca dll in favor of toolbox one
Toolbox tries to inject gwca dll to interface with the game, but more than 1 gwca dll running causes issues. If one already exists but isn't the right one we're looking for, we need to try to terminate that existing module using whatever exposed gwca functions are available on the already injected module, and then unload the old library before we inject the correct one.

3vcloud:
If a gwca.dll is already loaded in the process (e.g. from a previous toolbox run, a crash, or another GWCA-based tool), detect it via GetModuleHandleW, call GW::DisableHooks and GW::Terminate on it using GetProcAddress with the known mangled symbol names, then loop FreeLibrary until the module is fully unloaded before injecting the correct version.

Fixes #1945

### DIFF
--- a/GWToolboxdll/GWToolbox.cpp
+++ b/GWToolboxdll/GWToolbox.cpp
@@ -279,6 +279,29 @@ namespace {
                 && file_version_patch == GWCA::VersionPatch);
     }
 
+    void TerminateExistingGWCADll(HMODULE existing_gwca)
+    {
+        typedef void(__cdecl* GWFnVoid)();
+        const auto disable_hooks = (GWFnVoid)GetProcAddress(existing_gwca, "?DisableHooks@GW@@YAXXZ");
+        const auto terminate = (GWFnVoid)GetProcAddress(existing_gwca, "?Terminate@GW@@YAXXZ");
+        if (disable_hooks) {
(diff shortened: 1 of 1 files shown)


3vcloud:
@gunnarbeutner this might screw up your hot reload stuff, idk

gunnarbeutner:
This will almost certainly break other software that depends on a different GWCA DLL because this obviously doesn't patch up its IAT and just unloads the DLL right under its nose.

gunnarbeutner:
I think the real underlying problem is probably that GWCA doesn't handle the case properly where some of the functions it has hooked are already hooked through some other means (e.g. another instance of the GWCA DLL). Fixing that would be the right way if that's the problem. But that's just a guess because I don't have the GWCA source code.

3vcloud:
the hook are recursive, but rely on being unhooked in the same order, and also rely on the nested hook having enough space to recursively hook again. we came up against this a few years ago, but at the end of the day it doesn't matter - this is an exercise in gw not crashing due to other plugins and us debugging it

PR #1968, opened 2026-05-10
feat: editable outpost aliases in Travel Window

### LINKED ISSUE #1967: Ability to create your own aliases for maps in travel window module
@claude We have hard coded aliases for maps in the travel window e.g. gtob for great temple of balthazar, eee for embark beach english district 1. Instead, make this an editable list that would allow a user to remove or add new aliases for areas. The district and region are optional - if omitted it'll just use the current ones, just like the current logic does.

Add a reset button (use the confirm button functionality from out imgui code) that would reset the default aliases that come with toolbox.

While you're there, review the code for the travel window module to reduce LOC and remove unused code

3vcloud:
Implements #1967: editable map aliases in the Travel Window module.

- Aliases are now stored in a user-editable list (persisted to INI) instead of a hard-coded const map
- First run auto-populates from the built-in defaults (gtob, eee, doa, etc.)
- Settings panel shows an alias table: edit key, map, district, district number; add/delete rows
- "Reset to Defaults" button with ConfirmButton confirmation dialog
- Removed unused dungeon arrays, duplicate alias, stray character, and commented-out code

Closes #1967

### DIFF
--- a/GWToolboxdll/Windows/TravelWindowConstants.h
+++ b/GWToolboxdll/Windows/TravelWindowConstants.h
@@ -20,8 +20,8 @@ struct DistrictAlias {
     uint8_t district_number = 0;
 };
 
-// List of shorthand outpost names. This is checked for an exact match first.
-const std::map<const std::string, const OutpostAlias> shorthand_outpost_names = {
+// Default outpost aliases shipped with toolbox. Used for initialization and reset.
+const std::map<const std::string, const OutpostAlias> default_outpost_aliases = {
     {"bestarea", {GW::Constants::MapID::The_Deep}},
     {"toa", {GW::Constants::MapID::Temple_of_the_Ages}},
     {"doa", {GW::Constants::MapID::Domain_of_Anguish}},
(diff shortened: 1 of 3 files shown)


GWToolboxdll/Windows/TravelWindowConstants.h
```diff
@@ -76,40 +75,10 @@ constexpr std::array presearing_map_names = {
...
-    GW::Constants::MapID::Longeyes_Ledge_outpost,
-    GW::Constants::MapID::Sifhalla_outpost,
-    GW::Constants::MapID::Sifhalla_outpost,
-    GW::Constants::MapID::Olafstead_outpost,
-    GW::Constants::MapID::Umbral_Grotto_outpost,
-    GW::Constants::MapID::Gadds_Encampment_outpost,
-    GW::Constants::MapID::Deldrimor_War_Camp_outpost
-};
-
-constexpr std::array searchable_dungeon_names{
```
3vcloud, line 98: @claude why was this removed?

3vcloud: @claude the purpose of this code was originally to map a dungeon name to an outpost to easy travel - because dungeons aren't on the world map, the nearest outpost check wouldn't have worked. Since then we've added MapAdjacency.h to handle this, but it looks like it might need amending to add lower dungeon level mapping e.g. traversing from kathandrax level 3 > level 2 > level 1 > sacnoth > doomlore shrine to find its outpost.

3vcloud:
@claude instead of having fetched_searchable_outposts and the other one, using the map adjacency constants, build a single list of searchable areas that the player can travel to. This would allows us to search for dungeons, which isn't currently possible.

Add kathandrax to the list of default aliases too.

Identify any concerns with this e.g. missing areas that may have been in the original logic

PR #1970, opened 2026-05-10
Add BountyKillTracker widget

3vcloud:
Implements the Bounty Kill Tracker widget as described in issue #1383.

Overlays a kill counter on active bounty effect icons by tracking title point progress via `kTitleProgressUpdated`.

Closes #1383

Generated with [Claude Code](https://claude.ai/code)

PR #1972, opened 2026-05-11
Show bounty kill counter when count is 0

3vcloud:
Show the kill counter overlay even when the count is 0, so it appears immediately when a bounty is picked up.

References #1970


3vcloud:
@claude IsBountyTitle won't work - bounty skills don't directly map to their associated title. Instead, we need a GetBountyTitle(GW::Constants::SkillID) function containing a switch statement

PR #1973, opened 2026-05-11
feat: unify searchable areas list to enable dungeon travel search

3vcloud:
Merges the two separate searchable lists (outposts + explorable zones) into a single `searchable_areas` vector that also includes dungeon maps with adjacency data.

- `/tp catacombs of kathandrax` and other dungeon names now work
- `Travel()` auto-resolves non-outpost map IDs to nearest accessible outpost
- Adds `kathandrax` default alias
- Removes the second async decode loop (one less startup cycle)

Builds on #1968

### DIFF
--- a/GWToolboxdll/Windows/TravelWindow.cpp
+++ b/GWToolboxdll/Windows/TravelWindow.cpp
@@ -398,13 +398,12 @@ namespace {
             return bestMatchMapID;
         };
         auto best_match_map_id = GW::Constants::MapID::None;
-        if (ImInPresearing()) {
-            best_match_map_id = FindMatchingMap(compare.c_str(), presearing_map_names.data(), presearing_map_ids.data(), presearing_map_ids.size());
+        if (fetched_searchable_areas == FetchedMapNames::Ready) {
+            best_match_map_id = FindMatchingMapVec(compare.c_str(), searchable_areas);
         }
-        else {
-            if (best_match_map_id == GW::Constants::MapID::None && fetched_searchable_areas == FetchedMapNames::Ready) {
-                best_match_map_id = FindMatchingMapVec(compare.c_str(), searchable_areas);
-            }
+        else if (ImInPresearing()) {
+            // Fallback while searchable_areas is still being decoded
+            best_match_map_id = FindMatchingMap(compare.c_str(), presearing_map_names.data(), presearing_map_ids.data(), presearing_map_ids.size());
         }
 
         if (best_match_map_id != GW::Constants::MapID::None) {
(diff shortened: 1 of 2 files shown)


GWToolboxdll/Windows/TravelWindow.cpp
```diff
@@ -839,35 +826,24 @@ void TravelWindow::Update(const float)
...
-    // Dynamically generate a list of all outposts that the game has rather than storing another massive const array.
-    switch (fetched_searchable_outposts) {
+    // Dynamically build a unified searchable list: outposts, world-map explorable zones, and dungeons
+    // with adjacency data (so GetNearestOutpost() can resolve them to a travelable outpost).
+    switch (fetched_searchable_areas) {
         case FetchedMapNames::Pending: {
-            BuildSearchableAreas(searchable_outposts, [](const GW::Constants::MapID map_id, const GW::AreaInfo*) {
-                return IsValidOutpost(map_id) && !GW::Map::IsPreSearing(map_id);
+            BuildSearchableAreas(searchable_areas, [](const GW::Constants::MapID map_id, const GW::AreaInfo* map) {
+                if (GW::Map::IsPreSearing(map_id)) return false;
```
3vcloud: We might also be in pre searing

3vcloud: @claude sort this

3vcloud: @claude this is wrong, it should have been a single line edit and removal of the pre searing map names vector completely. GW::Map::GetIsPreSearing() without an argument would tell us whether the current char is in pre searing, and then GW::MAp::GetIsPreSEaring(map_id) == false would become GW::MAp::GetIsPreSEaring(map_id) == am_i_in_presearing

GWToolboxdll/Windows/TravelWindow.cpp
```diff
@@ -401,10 +401,6 @@ namespace {
         if (fetched_searchable_areas == FetchedMapNames::Ready) {
             best_match_map_id = FindMatchingMapVec(compare.c_str(), searchable_areas);
         }
-        else if (ImInPresearing()) {
```
3vcloud: @claude Can we now get rid of ImInPresearing?

GWToolboxdll/Windows/TravelWindow.cpp
```diff
@@ -829,12 +825,12 @@ void TravelWindow::Update(const float)
     // with adjacency data (so GetNearestOutpost() can resolve them to a travelable outpost).
     switch (fetched_searchable_areas) {
         case FetchedMapNames::Pending: {
-            BuildSearchableAreas(searchable_areas, [in_presearing = ImInPresearing()](const GW::Constants::MapID map_id, const GW::AreaInfo* map) {
-                if (GW::Map::IsPreSearing(map_id) != in_presearing) return false;
+            BuildSearchableAreas(searchable_areas, [am_i_in_presearing = GW::Map::IsPreSearing()](const GW::Constants::MapID map_id, const GW::AreaInfo* map) {
+                if (GW::Map::IsPreSearing(map_id) != am_i_in_presearing) return false;
```
3vcloud: @claude Fetching map names happens only once when toolbox loads - this pre searing check cant be done here because if the player switches from pre to post searing, the searching functionality wont work because it didnt fetch the names. The pre searing check needs to be elsewhere, where it will be checked at run time instead.

GWToolboxdll/Windows/TravelWindow.cpp
```diff
@@ -839,38 +815,41 @@ void TravelWindow::Update(const float)
...
-                return IsValidOutpost(map_id) && !GW::Map::IsPreSearing(map_id);
-            });
-            fetched_searchable_outposts = FetchedMapNames::Decoding;
-        }
-        break;
-        case FetchedMapNames::Decoding: {
-            if (CheckSearchableAreasDecoded(searchable_outposts)) {
-                fetched_searchable_outposts = FetchedMapNames::Ready;
+    // Rebuild the visible subset whenever the pre-searing state changes (or on first ready)
+    if (fetched_searchable_areas == FetchedMapNames::Ready) {
```
3vcloud, line 841: @claude visible_searchable_areas should be a pointer. We should have 2 vectors; one for post and one for pre maps. These vectors should be populated when fetching map names, and the only switch here would be toggling which vector our visible_searchable_areas points to.

PR #1975, opened 2026-05-11
feat: add custom effect management and spirit timer tracking

### LINKED ISSUE #1771: Feature Request: Additional Spirit Timers
The spirit timers available are currently for spirits that produce passive effects such as Edge of Extinction, Shelter, Life, etc. 

The request is to formulate a way to incorporate other spirits such as those that deal damage with their attacks (Bloodsong, Vampirism, etc); any other spirits that are commonly used. 

Would need to accommodate variable durations calculated by (at least) base skill properties, points in Spirit's attributes from build, active cons, and any other skill or environmental modifiers that are affecting the Spirit's lifespan at time of its spawn. 

Would the vanilla Minions window ('Y') provide data or a platform to help facilitate any of this function?

3vcloud:
Adds `GW::Effects::AddCustomEffect` and `RemoveCustomEffect` to ToolboxUtils for synthetic effect management, and amends EffectsMonitorWidget to optionally track spirits (Bloodsong, Vampirism) in earshot range as fake player effects.

Resolves #1771

### DIFF
--- a/GWToolboxdll/Utils/ToolboxUtils.cpp
+++ b/GWToolboxdll/Utils/ToolboxUtils.cpp
@@ -816,14 +816,6 @@ namespace GW {
                 if (remaining > 0)
                     memmove(&arr.m_buffer[i], &arr.m_buffer[i + 1], remaining * sizeof(*arr.m_buffer));
                 arr.m_size--;
-                arr.m_capacity--;
-                if (arr.m_size == 0) {
-                    GW::MemoryMgr::MemFree(arr.m_buffer);
-                    arr.m_buffer = nullptr;
-                } else {
-                    auto* new_buf = static_cast<GW::Effect*>(GW::MemoryMgr::MemRealloc(arr.m_buffer, arr.m_size * sizeof(*arr.m_buffer)));
-                    if (new_buf) arr.m_buffer = new_buf;
-                }
                 GW::UI::SendUIMessage(GW::UI::UIMessage::kEffectRemove, reinterpret_cast<void*>(static_cast<uintptr_t>(effect_id)));
                 return true;
             }
--- a/GWToolboxdll/Utils/ToolboxUtils.cpp
+++ b/GWToolboxdll/Utils/ToolboxUtils.cpp
@@ -789,7 +789,7 @@ namespace GW {
 
             GW::UI::UIPacket::kEffectAdd packet{};
             packet.agent_id = GW::Agents::GetControlledCharacterId();
-            packet.effect = ToolboxUtils::GwArrayPushBack(arr, new_effect);
+            packet.effect = GW::MemoryMgr::AddToGuildWarsArray(arr, new_effect);
             GW::UI::SendUIMessage(GW::UI::UIMessage::kEffectAdd, &packet);
             return new_effect.effect_id;
         }
@@ -802,7 +802,7 @@ namespace GW {
             auto& arr = player_effects->effects;
             for (uint32_t i = 0; i < arr.m_size; i++) {
                 if (arr.m_buffer[i].effect_id != effect_id) continue;
-                ToolboxUtils::GwArrayErase(arr, i);
+                GW::MemoryMgr::RemoveFromGwArray(arr, i);
                 GW::UI::SendUIMessage(GW::UI::UIMessage::kEffectRemove, reinterpret_cast<void*>(static_cast<uintptr_t>(effect_id)));
                 return true;
             }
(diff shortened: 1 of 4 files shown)


GWToolboxdll/Utils/ToolboxUtils.cpp
```diff
@@ -754,6 +755,70 @@ namespace GW {
...
+
+            GW::Effect new_effect{};
+            new_effect.skill_id = skill_id;
+            new_effect.effect_id = next_custom_effect_id++;
+            new_effect.duration = duration_seconds;
+            new_effect.timestamp = GW::MemoryMgr::GetSkillTimer();
+            new_effect.attribute_level = 0;
+            new_effect.agent_id = 0;
+
+            arr.m_buffer[arr.m_size] = new_effect;
```
3vcloud: @claude We don't have control over the memory in the player effects array - its managed by gw, so the current functionality is UB. We've exposed GW::MemoryMgr::Realloc to handle this though - check the quest module for an example of how we've previously grown the array to fit. Same situation when removing the effect - use GW::MemoryMgr::Free, and check the quest module for an example of that, too

GWToolboxdll/Widgets/EffectsMonitorWidget.cpp
```diff
@@ -166,11 +216,18 @@ void EffectsMonitorWidget::Initialize()
...
 {
     ToolboxWidget::Terminate();
     GW::UI::RemoveUIMessageCallback(&OnPreUIMessage_HookEntry);
+    GW::StoC::RemoveCallback<GW::Packet::StoC::AgentRemove>(&OnAgentRemove_SpiritEntry);
+    for (const auto& [agent_id, effect_id] : tracked_spirit_effects) {
+        GW::Effects::RemoveCustomEffect(effect_id);
+    }
+    tracked_spirit_effects.clear();
 }
 void EffectsMonitorWidget::Update(float)
```
3vcloud: @claude the logic in the update loop doesn't have to happen every single frame - use the delta float given to limit it to ~30fps. Because gw runs the update and draw loop in the same thread, the idea is to reduce the amount of stuff it needs to do in a single frame.

Reduce LOC in this function too - combine the if statements if you can, use GW::Agents::GetAgentMatchesFlags for the alive spirit check, then you can safely cast to AgentLiving without the extra lines

GWToolboxdll/Utils/ToolboxUtils.cpp
3vcloud: @claude dont free the buffer or reduce capacity, just reduce the size and shift the other elements as needed. Any new effects coming in will be able to then be added to the array without having to reallocate data because capacity will fit.

Should this logic be in ToolboxUtils instead, now that it'll be used on 2 places?

3vcloud: @claude Add GW::MemoryMgr functions un toolboxutils to add to or remove from an array in gw memory

GWToolboxdll/Utils/ToolboxUtils.cpp
```diff
@@ -754,6 +755,62 @@ namespace GW {
...
+            new_effect.skill_id = skill_id;
+            new_effect.effect_id = next_custom_effect_id++;
+            new_effect.duration = duration_seconds;
+            new_effect.timestamp = GW::MemoryMgr::GetSkillTimer();
+            new_effect.attribute_level = 0;
+            new_effect.agent_id = 0;
+
+            GW::UI::UIPacket::kEffectAdd packet{};
+            packet.agent_id = GW::Agents::GetControlledCharacterId();
+            packet.effect = ToolboxUtils::GwArrayPushBack(arr, new_effect);
```
3vcloud: @claude this should be GW::MemoryMgr::AddToGuildWarsArray, and similarly refactor for GW::MemoryMgr::RemoveFromGwArray

GWToolboxdll/Utils/ToolboxUtils.cpp
```diff
@@ -754,6 +755,62 @@ namespace GW {
...
+            GW::UI::SendUIMessage(GW::UI::UIMessage::kEffectAdd, &packet);
+            return new_effect.effect_id;
+        }
+
+        bool RemoveCustomEffect(const uint32_t effect_id)
+        {
+            if (effect_id < custom_effect_id_start) return false;
```
3vcloud, line 799: @claude this logic would just keep incrementing the available effect ids we can use - we need a pool so we can reuse ids

PR #1981, opened 2026-05-11
Travel Window: all buttons in 2-col grid, user-configurable destinations

### LINKED ISSUE #1409: Travel Window, more customization possible?
- Would it be possible to add an option to the "Travel" Window Settings to not display the 8 "default destinations"? (I'm honestly not using them much besides "Embark Beach" and would rather not have them there.)
- Would it be possible to have the personally added destinations in a "2 by x grid", similar to how the default destinations are arranged? (Since I'm not adding and removing them constantly for me it would be fine if that was done though a separate menu, keeping the one constantly used cleaner)
- Would it be possible to have campaign identifiers next to the destination names?
- I also strongly support issue #1368 

(I hope it's ok to bundle these here. Thanks :D )

3vcloud:
Revisits the Travel Window as requested in #1409.

## Changes

- All outpost buttons now render in a 2-column grid (the half-width layout previously used only for the 8 hard-coded defaults now applies to all user destinations)
- The 8 hard-coded travel buttons become the default pool for a new `user_destinations` vector; users can add, remove and reset these in settings
- The inline favourites editor (dropdown + arrows + edit pen) is removed from the main window — all destination management moves to a new "User Travel Destinations" table in settings
- New "Show Zaishen quest buttons" checkbox replaces the old "Show default destinations" toggle
- Backwards-compatible migration: reads old `fav_count`/`Fav%d` keys on first run

Closes #1409

### DIFF
--- a/GWToolboxdll/Windows/TravelWindow.cpp
+++ b/GWToolboxdll/Windows/TravelWindow.cpp
@@ -221,19 +221,25 @@ namespace {
     GW::HookEntry OnUIMessage_HookEntry;
 
 
-    const std::vector<GW::Constants::MapID> default_user_destinations = {
-        GW::Constants::MapID::Temple_of_the_Ages,
-        GW::Constants::MapID::Domain_of_Anguish,
-        GW::Constants::MapID::Kamadan_Jewel_of_Istan_outpost,
-        GW::Constants::MapID::Embark_Beach,
-        GW::Constants::MapID::Vloxs_Falls,
-        GW::Constants::MapID::Gadds_Encampment_outpost,
-        GW::Constants::MapID::Urgozs_Warren,
-        GW::Constants::MapID::The_Deep,
+    struct UserDestEntry {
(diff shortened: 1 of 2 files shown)


GWToolboxdll/Windows/TravelWindow.cpp
```diff
@@ -221,13 +221,29 @@ namespace {
     GW::HookEntry OnUIMessage_HookEntry;
 
 
-    // ==== Favorites ====
-    std::vector<GW::Constants::MapID> favourites{};
+    const std::vector<GW::Constants::MapID> default_user_destinations = {
```
3vcloud: @claude I asked for this feature to be similar in functionality to the /tp aliases, but you havent added district options

3vcloud: @claude tackle this

3vcloud:
@claude I asked for this feature to be similar in functionality to the /tp aliases, but you havent added district options. Amend the logic to include this,

PR #2001, opened 2026-05-13
"Splits" window for tracking different goals for speed runners.

jdm080:
A ways to go, but wanted to see if this is even something that might be wanted.

Adds a new built-in ToolboxWindow for tracking different speedrun's and their splits in Guild Wars. Features include:

- Split timer with real time and game time tracking
- Goal-based triggers: mission complete/bonus, map enter/exit, vanquish complete, reach level, reach title rank, manual
- Hard mode flag on mission/bonus triggers
- Batch goal pickers for missions, explorables, towns, and titles
- Personal best comparison (real or game time basis)
- Per-character run history saved to JSON
- Crash-protection resume on toolbox restart
- Configurable hotkeys for start, reset, and manual split
- Settings panel integrated into the standard Toolbox settings UI

image

image

image


3vcloud:
@claude between this and the existing objective timer module, how would you approach having a single source of truth for tracking objectives in game without impacting the look and feel of the existing objective timer in toolbox? Could we enhance this version with what's missing to be a replacement of the existing module? We also already have a feature to host a websocket to host live splits, could this easily be integrated?

I'm conscious of how much code this module would add, and would like to merge functionality related to objective tracking

jdm080:
So as an overall direction, more integration with the current objective timer (full collapse into or via a shared bus). Webserver integration also.

3vcloud:
> So as an overall direction, more integration with the current objective timer (full collapse into or via a shared bus). Webserver integration also. 

Yeh looks like it's a close fit to replace the objective timer to be honest, so with a few more tweaks I wouldn't have a problem using this instead

jdm080:
Left OT as its own Window(for now) and all its current saving behavior/looks etc, instead both splits and OT just subscribe to the "GwEventBus". 2 profiles in Splits Manual/Running. Cut down a few settings/added a few different ones.
Splits sends commands to web socket server when its enabled.  Added a section in the websocket server. Just so people can know why a command (start/reset/split/etc) was sent and from what window+why.
image
image
image
Screenshot 2026-07-11 095559

Happy to make any further tweaks before\after merge.
(Fully integrate current OT into the splits window as a profile as an example)

jdm080:
Progressed farther so bringing back to draft ( OT/Web socket still there but once splits is validated its basically ready to go)

SC profile implemented = OT (Validating still lots of stuff to verify)
SC still has its default files, but allows you to create a duplicate list and name it example being  2 man FoW. so you have the generic bucket for people who want the old behavior but adds the ability to section it via different save files.

Quest+Skill added (no internal db just made a button to /wiki game integrations to get the quest id or skill ID you want from there.
Mob Kill goal. 
Auto fail on rezone
Comparison toggle for PB/AVG/Last Run

Nuzlocke section
     Tracks party deaths player+hench+hero 
     Points section
added some helpful texts in sections to better explain some stuff.

3vcloud:
@claude theres a lot of code in this PR that is duplicated logic e.g. decoding strings, getting map names, loading settings. The gw event bus thing needs to go; hooks should be done from within the module, not via another. Theres too much StoC handling going on to the point that GWCA edits are suggested; many of these hooks have UIMessage equivalents - give examples such as the quest one, and identify others. There are functions like "GetNByString" or "GetStringByN" which should really just be maps to reduce LOC, and many other examples of over engineering to do a job that would be easier and still maintainable with much less code. The UI layout in the settings is a bit confusing, but I'm not sure how much of a help AI can be with this bit.

Amends have been (accidentally) added to the PR that would affect CI, and these need to go. Agents.md needs consulting against this code aswell.

Its been created using AI, but if it works its a really comprehensive solution - its just too much code for what it does, and it could do the same thing with much less.

Summarise this to provide constructive feedback, using the existing codebase as a reference

jdm080:
### Cleanup pass: LOC reduction from review feedback

In response to review feedback on duplication/over-engineering, went through and consolidated the genuine duplicates (not a blanket rewrite — verified each one is actually the same logic before touching it). Net **-1,047 lines** across all Splits/OT/WebSocketModule files since the last feature-build commit (`243e2fc7`), via `git diff --numstat`:

| File | Net lines | What changed |
|---|---|---|
| `SplitsGoalListWindow.cpp` | −349 | Dynamic title list, trigger-type table, `DrawSettings` split into one method per profile/section, Town/Explorable batch pickers merged into one parameterized picker |
| `SplitsWindow.h` | −106 | NuzlockeState struct extraction |
| `GWEventBus.cpp/.h` | −443 (deleted) | Replaced with direct hooks in OT/SplitsWindow — architecture change, not compression |
| `MapNames.cpp/.h` | −72 (deleted) | Dead code, superseded by `Resources::GetMapName` |
| `GoalEngine.cpp` | −68 | Dead `NotifyMissionBonus` removed, `ReachTitleRank` simplified |
| `GoalEntry.h` | −40 | |
| `GoalList.cpp` | −16 | Enum↔string table consolidation |
| `SplitsWindow.cpp` | −14 | Near-neutral (hook additions offset helper extractions) |
| `ObjectiveTimerWindow.cpp` | +77 | Grew — absorbed the hooks GWEventBus used to own |

Honest caveat: of the −1,047, roughly half is genuine duplicate/dead-code removal; the rest is the GWEventBus architecture change (relocated logic, not eliminated) and the NuzlockeState reorganization. Not claiming this hits any specific target percentage — just what held up under actually reading the code.

I'll dig a little deeper tomorrow.

henderkes:
ok but could you perhaps write anything on your own instead of just prompting the slop machine? use it for code all you want but don't subject us to reading that.

there's also still a merge conflict

3vcloud:
I'm on the phone atm but I can see it hasn't moved the majority of stoc callbacks into ui messages. the websocket module is actually just another part of the objective timer at this point and not a separate module on its own, unless it's a utility that more than 1 module can use but then it still not a module.

Will take a look this evening, thanks

jdm080:
OT/  the original Websocket is not deleted/removed yet. The end goal was to replace it fully. (Do I just fully delete it right now?) was keeping to compare functionality.

image

As for the StoC calls 11 orginally were copies from OT itself.
ManipulateMapObject,InstanceLoadInfo,InstanceLoadFile,GameSrvTransfer,MessageServer,DisplayDialogue,,AgentUpdateAllegiance,DoACompleteZone,CountdownStart are the ones left now. (don't seem to be any replacements)

Splits StoC calls
PartyLock - checking for the mission-que to normalize timer

5 for the Nuzlocke section
There isn't any place to get Henchman name in GWCA besides PartyHenchmanAdd?
Which is used for getting the names into the roster, it would be possible to swap over and have split between StoC for the 1 call. Other 4 could use player/hero add/remove but wouldn't be fully enough to remove StoC so kept for consistency with itself.

image

3vcloud:
@claude is there anything in the toolbox codebase that programmatically gets a hero name?

3vcloud:
@claude finally, is there anything in the toolbox codebase that can get a henchman name, either in an explorable area or in an outpost?

3vcloud:
ok claude was wrong, but you'd get the henchman name by agent id. I did write somewhere how to find the list of henchmen agent ids available in an outpost, but not sure if thats needed

GWToolboxdll/Modules/ToolboxSettings.cpp
```diff
@@ -263,7 +263,6 @@ namespace {
         {AccountInventoryWindow::Instance(), false},
         {PerformanceWindow::Instance(), false},
         {PacketLoggerWindow::Instance(), false},
-        TexmodModule::Instance()
```
henderkes: hm?

jdm080: Line still there! Two merges of dev happened. Both added the same line. Duplicate got removed.

jdm080:
All 3 profiles in, Manual/Running/SC(OT replacement hence a lot of duplication currently) 

Found some replacements for some of the current OT hooks for "Splits" to use. UImessage equivalents..

Objective started: OT → ObjectiveUpdateName (StoC), Splits → kObjectiveUpdated (UI message)
Objective done: OT → ObjectiveDone (StoC), Splits → kObjectiveComplete (UI message)
Dungeon reward: OT → DungeonReward (StoC), Splits → kDungeonComplete (UI message)

Added in a few behaviors to let SC function more like OT. (auto starts/resets after failing/completing). Presets folder so they don't clutter the user list.   Cleaner to not show every file and instead just show user created ones and if a user never creates their own it just uses the default and functions similarly to current OT.

Nuzlocke rewritten. diff's against GetPartyInfo instead of the previous StoC.

This has gotten way larger than anticipated . So wanted to ask for some clarity or let me know whatever you guys want.
1. Workable as is for a PR.
2. Reworked to be multiple smaller PRs over time.
3. Make it a plugin instead.

PR #2051, opened 2026-05-18
Add copy-to-clipboard buttons for build and teambuild codes

### LINKED ISSUE #2047: Move all build load logic into a static Build::Update() function
@claude When a player wants to load a build, no matter if its from the builds window module or from the hero builds module, it should just have to call Build::Load() - I've added this stub with a todo comment.

Inside the Load() function, the build should be copied into a vector in private namespace in the teambuilds cpp file. A static Build::Update() function will be called every frame, and will loop this vector to update the build(s) being loaded in, be it waiting for a hero to be loaded or anything else.

This should centralise all of the async build loading logic in one tidy function, and futher simplify the hero builds and builds windows. The Build::Update() function can be called from the GWToolbox cpp file in its update loop.

The end result should allow a build.Load() to be called anywhere, regardless of the origin.

I've also revisited the Send() functions, but I've not added a send_queue back in - this queue was originally there to make sure we don't send too many messages at once, so this should also be cycled in the Build::Update() function

3vcloud:
Closes #2047

Adds copy-to-clipboard functionality for both individual builds and team builds:

- `Build::Copy()` copies `[name;code]` to clipboard
- `TeamBuild::Copy()` copies `[TB;<encoded>]` to clipboard
- Player builds layout: small copy icon button between Edit and Delete
- Hero builds layout: copy icon button after Load on each row
- Edit window footer: "Copy Teambuild code" button next to "Send Teambuild code in chat"

### DIFF
--- a/GWToolboxdll/Utils/TeamBuild.cpp
+++ b/GWToolboxdll/Utils/TeamBuild.cpp
@@ -340,6 +340,21 @@ void Build::Send() const
     EnqueueSend(msg);
 }
 
+void Build::Copy() const
+{
+    if (code.empty() && name.empty()) return;
+    auto gen_name = name;
+    if (hero_id != GW::Constants::HeroID::NoHero) {
+        if (gen_name.empty())
+            gen_name = Resources::GetHeroName(hero_id)->string();
+        else
+            gen_name = std::format("{} ({})", name, Resources::GetHeroName(hero_id)->string());
+    }
+    const auto msg = code.empty() ? name : std::format("[{};{}]", gen_name, code);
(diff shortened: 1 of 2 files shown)


3vcloud:
@claude teambuilds can now have any number of player builds. When we click on a teambuild in chat it opens the teambuild window expecting it to be 1 player and up to 7 heros - instead it might be a teambuild designed for a team of up to 12 players, possibly without any heros at all. Amend the logic to gracefully handle teambuilds that are designed for players; allow the teambuild to be saved to the builds window module as applicable, and theres no reason for a "load all" button unless its a hero loadout.

PR #2056, opened 2026-05-19
Disable send-to-chat button when encoded teambuild exceeds 120 chars

### LINKED ISSUE #2054: Disable sending encoded teambuild to chat if its too long
@claude Depending on the teambuild, the code generated for chat could be too long for a chat message. If the code would be too long ( [TB;<code>] should be less than 120 chars), visually disable the button and provide a tooltip to explain why.

3vcloud:
Closes #2054

Adds `TeamBuild::ChatCodeTooLong()` which encodes the teambuild and checks whether `[TB;<encoded>]` would reach the 120-wchar GW chat limit.

Two send buttons are guarded:
- "Send Teambuild code in chat" in `DrawEditWindow` (TeamBuild.cpp)
- Ctrl+click "Send" in `HeroBuildsWindow::draw_teambuild_row`

Both use `ImGui::BeginDisabled/EndDisabled` for visual feedback and show an explanatory tooltip via `ImGuiHoveredFlags_AllowWhenDisabled`.

### DIFF
--- a/GWToolboxdll/Utils/TeamBuild.cpp
+++ b/GWToolboxdll/Utils/TeamBuild.cpp
@@ -428,6 +428,18 @@ void TeamBuild::SetSkillToggleSprite(IDirect3DTexture9** sprite)
     skill_toggle_sprite = sprite;
 }
 
+const std::wstring& TeamBuild::GetEncoded() const
+{
+    if (!encoded_cache_.has_value())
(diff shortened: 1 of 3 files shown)


3vcloud:
@claude encoding/decoding is expensive; can we implement some sort of cache within a build/teambuild to avoid reprocessing it?

PR #2057, opened 2026-05-19
Show item quantity totals in Account Inventory tooltip

gunnarbeutner:


3vcloud:
@claude the on item tooltip call is made every frame. Can you suggest a refactor to make it less expensive without adding too much code?

3vcloud:
@claude action your suggestion. The idea is to have the tooltip logic as lightweight as possible

3vcloud:
@claude what you've done is overkill; there will only be one tooltip in the app so we can use static variables inside the tooltip function without assigning extra memory to other structs. Compare the pointer passed to the function - we know that we have to recalculate the tooltip variables if it's changed.

PR #2058, opened 2026-05-19
Fix IsHeroUnlocked check for mercenary heroes

gunnarbeutner:
The previous condition (a.name && !a.name) was a tautological contradiction that always evaluated to false, causing IsHeroUnlocked to always return false for mercenary heroes and blocking them from being added when loading a team build. HeroInfo::name is a fixed wchar_t[20] array, so check the first character to detect an unassigned merc slot.


3vcloud:
Nice catch, thanks

PR #2060, opened 2026-05-19
Add mission map icons renderer to MissionMapWidget

### LINKED ISSUE #1836: Resurection Shrines on Misssion Map (U Map)
I can change the U maps icons and colors of enemies and other entities , But resurrection shrines(specifically for Nightfall or eotn and what blessing they provide,) I cannot change, id love to see what blessings they provide, but if i could JUST get coloring them something more noticeable on U Map that would achieve my needs 

I believe Blesssing locations would probs be a texmod concern, But the Mission Map icon itself is a Toolbox concern. If anyone could point me in right place for a texmod request that would be gr8

*This is less of an issue and more of a request- I was told to post here from discord.

3vcloud:
Adds a few-line renderer to `MissionMapWidget::Draw` that draws each entry from `WorldContext::mission_map_icons` as its actual in-game texture on the mission map, centred at its world-map X/Y position and clipped to the mission map bounds.

Closes #1836

### DIFF
--- a/GWToolboxdll/Widgets/MissionMapWidget.cpp
+++ b/GWToolboxdll/Widgets/MissionMapWidget.cpp
@@ -1,6 +1,7 @@
 #include "stdafx.h"
 
 #include <GWCA/Context/GameplayContext.h>
+#include <GWCA/Context/WorldContext.h>
 #include <GWCA/GameEntities/Agent.h>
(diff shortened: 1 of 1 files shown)


3vcloud:
@claude re-introduce the colour alteration so that the icon colours behave the same way gw handles them

PR #2061, opened 2026-05-19
Add affiliation colour tints to mission map icons

3vcloud:
Re-introduces colour alteration for `WorldContext::mission_map_icons` so that each icon is tinted by its `option` field (affiliation: gray, blue, red, yellow, teal, purple, green), matching GW's rendering.

Follows up on #2060.


3vcloud:
@claude this method of tinting is wrong; the colour balance of the icon isn't applied in the same way the game does it, so you must be handling the tint wrongly

PR #2062, opened 2026-05-19
Fix mission map icon colour tinting (use SELECTARG2 not MODULATE)

3vcloud:
Fix mission map icon colour tinting to use SELECTARG2 instead of MODULATE.

GW fills icons with a solid affiliation colour, using the texture only as an alpha mask. The previous AddImage tint approach used D3DTOP_MODULATE which multiplied every texture RGB channel by the tint — distorting the colour balance. An AddCallback pair now switches to SELECTARG2/DIFFUSE around the icon draw commands to match GW's rendering.

Follows up on #2061.


3vcloud:
@claude still not the right approach. We need to desaturate the image and then apply the tint, and we probably need to use a pixel shader to do that. Create and apply a pixel shader that will apply our tint in this way to see if that sorts it out

PR #2070, opened 2026-05-21
Add mobile/normal mode layout settings to UI elements

3vcloud:
Implements #3 2069: separate size/position settings for mobile and non-mobile game views.

- Tracks `GW::UI::IsInMobileMode()` each frame via `ToolboxSettings::is_in_mobile_mode`
- Adds mobile-specific layout fields (lock_move, lock_size, auto_size, snapped_frame_label, pos, size)
- `DrawSizeAndPositionSettings` now shows Normal/Mobile vertical tabs on the left with per-mode settings on the right
- All settings elements are always drawn; inapplicable ones use BeginDisabled() + tooltip instead of being hidden
- On mode change, window positions/sizes are automatically saved and restored

### DIFF
--- a/GWToolboxdll/Modules/ToolboxSettings.cpp
+++ b/GWToolboxdll/Modules/ToolboxSettings.cpp
@@ -5,6 +5,7 @@
 #include <GWCA/Managers/MapMgr.h>
 #include <GWCA/Managers/AgentMgr.h>
 #include <GWCA/Managers/GameThreadMgr.h>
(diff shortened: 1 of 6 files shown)


3vcloud:
@claude the tab button text should be vertically oriented and it should be more obvious as to what content is within the tab. We'll be using this style of tabbing data a lot (e.g. vertically oriented tabs on the left, tab content on the right - can you come up with something to add to ImGuiAddons that would easily allow us to create tabbed content in this format?

PR #2078, opened 2026-05-22
fix: hotkey group children lost on reload due to alphabetical INI sec…

### LINKED ISSUE #2077: Hotkey groups are completely broken
- New hotkeys created under groups are not properly saved and loaded
- Pre-8.18 hotkeys stored in groups will be converted to the new format, but I've already had this system fail on me, and it ended up deleting all of them. Dunno how to reproduce, unfortunately.

henderkes:
…tion ordering

CSimpleIni returns sections in alphabetical order. Child sections like "hotkey-000-001:SendChat" sort before their parent "hotkey-000:HotkeyGroup" because '-' (ASCII 45) < ':' (ASCII 58). The previous single-pass loader called groups_by_id.find() for each child before the parent group section had been processed, so every child hotkey was silently discarded on load.

Fix: split the loader into a genuine two-pass approach - pass 1 loads all top-level hotkeys/groups (populating groups_by_id), pass 2 attaches child sections to their now-known parent groups.

Fixes #2077

### DIFF
--- a/GWToolboxdll/Windows/HotkeysWindow.cpp
+++ b/GWToolboxdll/Windows/HotkeysWindow.cpp
@@ -615,52 +615,62 @@ void HotkeysWindow::LoadSettings(ToolboxIni* ini)
     ini->GetAllSections(entries);
     bool had_migration = false;
 
-    // Two-pass: first collect group objects keyed by their numeric ID so we can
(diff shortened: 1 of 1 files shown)


henderkes:
@claude look at using std::string_view instead of string

PR #2085, opened 2026-05-23
Fix disconnect when selecting custom map marker quest

### LINKED ISSUE #1917: Creating a custom marker crashes TB
GW::QuestMgr::SetActiveQuestId causes the issue, related to the quest entry not yet being in the quest log frame

JulienTant:
Fixes #1917.

The custom map marker is represented as a local synthetic quest. Selecting it must not send quest operations to the server, because the server does not know that quest id.

This patch prevents two remaining leak paths:
- `ParseQuestObjectives()` no longer requests missing quest info for the custom marker quest.
- `SetActiveQuestId()` forces custom marker selection to be local-only.
- The world map context menu now routes "Set active quest" through `QuestModule::SetActiveQuestId()` instead of calling `GW::QuestMgr::SetActiveQuestId()` directly.

Manual test:
- Built `GWToolboxdll.dll` from this branch (using GitHub Actions on my fork because I'm on linux haha).
- Replaced the DLL.
- Created and selected a custom marker.
- Confirmed no disconnect.

Note: this patch was AI-generated and manually tested before opening the PR.


henderkes:
thank you

PR #2111, opened 2026-05-26
Add "No Build Defined" placeholder in TeamBuild hover tooltip

### LINKED ISSUE #2110: Hero Builds flyout change proposal.
The new Hero Builds summary flyout (v useful, TY) does not show Heroes that have no build code defined.

It would be useful to see all names in the team with a red warning text like 'No Build Defined' instead of the skill icons for those affected.

3vcloud:
Closes #2110

Adds `TeamBuild::DrawTooltip()` which consolidates the hover-tooltip logic from both HeroBuildsWindow and BuildsWindow into the TeamBuild class.

Heroes with no build code (or an invalid code) now show a red "No Build Defined" placeholder in the tooltip instead of being silently omitted.

### DIFF
--- a/GWToolboxdll/Utils/TeamBuild.cpp
+++ b/GWToolboxdll/Utils/TeamBuild.cpp
@@ -470,6 +470,7 @@ const std::wstring& TeamBuild::GetEncoded() const
 void TeamBuild::ResetEncodedCache() const
 {
     encoded_cache_.reset();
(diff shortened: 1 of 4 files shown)


3vcloud:
@claude is there a way we could cache the logic in DrawTooltip to avoid having to build it every frame?

PR #2112, opened 2026-05-26
Notepad: only show close button on active tab

### LINKED ISSUE #2109: Notes text cleared after version X.XX
Version History does not record when Notes changed but the change to having a tabbed facility appears to have cleared previously stored text. 

I am requesting a way to recover that text, which is important to me.

3vcloud:
Only show the X (close) button on the currently active Notepad tab to prevent accidental closure of short-named tabs.

Inactive tabs no longer show the X, but tabs can still be closed via right-click → Close.

Closes #2109


3vcloud:
@claude trigger the confirm prompt when we try to remove a tab; we've got code in imguiaddons to handle confirm dialog

PR #2133, opened 2026-05-28
fix: MaterialsWindow use kVendorTransComplete as primary transaction signal

### LINKED ISSUE #2131: Materials module buys/sells items too quickly when on low ping, triggers anti-flood
The materials module attempts to buy/sell items too quickly when playing with low ping(<30ms), causing it to hit the flood/macro limit after 12 trades. This causes it to stall for several seconds before sending a "Timeout waiting for transaction" error. After this I'd have to manually interact with the trader again, otherwise any further attempts to use the module to buy/sell would stall with a "Timeout waiting for quote" error.

A solution would be to set a short delay after succesfully buying/selling something, as it doesn't happen when my ping is higher than about 35-40ms.

3vcloud:
Fixes #2131

Listen for `kVendorTransComplete` to detect when the server has finished processing a trade, preventing the next quote from being sent before the server closes the previous transaction. This fixes the race condition on low-ping connections caused by a recent GW server packet ordering change.

The item count change check is kept as a secondary fallback signal. The unused `MIN_TIME_BETWEEN_RETRY` constant is removed.

### DIFF
--- a/GWToolboxdll/Windows/MaterialsWindow.cpp
+++ b/GWToolboxdll/Windows/MaterialsWindow.cpp
@@ -402,12 +402,8 @@ void MaterialsWindow::Update(const float)
         trans->state = Transaction::State::Transacting;
     } break;
     case Transaction::State::Transacting: {
-        if (trans_complete) {
+        if (trans_complete && CountItemByMaterialSlot(trans->material) != trans->initial_item_count) {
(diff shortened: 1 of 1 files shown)


3vcloud:
@claude both checks need to pass, not one or the other

PR #2140, opened 2026-05-28
 Show Enemy Health Without Needing Player Damage

careyourcurroundings:
- Introduced `HealthLogCache` as the single shared loader for `healthlog.ini`
- `HealthLogCache` uses previously fixed `FastIni::GetAllKeys` so it returns key names correctly
- Updated `HealthWidget` and `PartyDamage` to use `HealthLogCache`
- Allow `HealthWidget` to display enemy HP/Max HP without needing player damage first, by loading Max HP value from `HealthLogCache`


3vcloud:
looks good, a few things:

Doesn't need to be its own module, should be part of damage monitor. If we want to grab max hp from elsewhere later in life, we'll just expose some static funcs in the class.

Don't give the option of loading from different places; we try to keep toolbox settings in the same contained place. I like the idea of keeping it a separate ini though.

What about HM/NM?

careyourcurroundings:
> looks good, a few things:
> 
> Doesn't need to be its own module, should be part of damage monitor. If we want to grab max hp from elsewhere later in life, we'll just expose some static funcs in the class.
> 
> Don't give the option of loading from different places; we try to keep toolbox settings in the same contained place. I like the idea of keeping it a separate ini though.
> 
> What about HM/NM?

Part of damage monitor sounds good.

I forgot that NM existed. I believe the difference is always lvl difference\*20. The easiest solution would be to calculate record the "base hp" and then apply level\*20 on top of it to get max hp. Do you think I should make a new ini and purge the old one, so ppl don't have stale numbers?

3vcloud:
> > looks good, a few things:
> > Doesn't need to be its own module, should be part of damage monitor. If we want to grab max hp from elsewhere later in life, we'll just expose some static funcs in the class.
> > Don't give the option of loading from different places; we try to keep toolbox settings in the same contained place. I like the idea of keeping it a separate ini though.
> > What about HM/NM?
> 
> Part of damage monitor sounds good.
> 
> I forgot that NM existed. I believe the difference is always lvl difference*20. The easiest solution would be to calculate record the "base hp" and then apply level*20 on top of it to get max hp. Do you think I should make a new ini and purge the old one, so ppl don't have stale numbers?

Same ini, key by mode and npc id, no calcs needed

careyourcurroundings:
> > > looks good, a few things:
> > > Doesn't need to be its own module, should be part of damage monitor. If we want to grab max hp from elsewhere later in life, we'll just expose some static funcs in the class.
> > > Don't give the option of loading from different places; we try to keep toolbox settings in the same contained place. I like the idea of keeping it a separate ini though.
> > > What about HM/NM?
> > 
> > 
> > Part of damage monitor sounds good.
> > I forgot that NM existed. I believe the difference is always lvl difference_20. The easiest solution would be to calculate record the "base hp" and then apply level_20 on top of it to get max hp. Do you think I should make a new ini and purge the old one, so ppl don't have stale numbers?
> 
> Same ini, key by mode and npc id, no calcs needed

Looks good? Should legacy loading be removed?

3vcloud:
Looks better, but still too complicated:

1. Put 2 maps in memory - one for hm, one for nm - no runtime hashing of strings needed
2. Put 2 sections in the ini - one for hm, one for nm - key value pairs then become npc id > hp
3. No need for boolean return value when trying to get cached health - 0 is an invalid result anyway, and tells us that for whatever reason the result of the call didn't work.
4. Get rid of the forward calling IsCurrentHardMode
5. No need for the runtime load on demand stuff, just load and save in the existing LoadSettings/SaveSettings calls as normal
6. Re: legacy loading - there is no legacy, so no need for legacy loading ;)

This is a good addition to the module, but its still too much code for what you're wanting to achieve. To implement this should be maybe 20 lines of extra code to the existing module.

careyourcurroundings:
> Looks better, but still too complicated:
> 
>     1. Put 2 maps in memory - one for hm, one for nm - no runtime hashing of strings needed
> 
>     2. Put 2 sections in the ini - one for hm, one for nm - key value pairs then become npc id > hp
> 
>     3. No need for boolean return value when trying to get cached health - 0 is an invalid result anyway, and tells us that for whatever reason the result of the call didn't work.
> 
>     4. Get rid of the forward calling IsCurrentHardMode
> 
>     5. No need for the runtime load on demand stuff, just load and save in the existing LoadSettings/SaveSettings calls as normal
> 
>     6. Re: legacy loading - there is no legacy, so no need for legacy loading ;)
> 
> 
> This is a good addition to the module, but its still too much code for what you're wanting to achieve. To implement this should be maybe 20 lines of extra code to the existing module.

Oh man I thought I read 20 lines of new code, not extra code. I was going to ask for more hints on how to get it down.

3vcloud:
If you'd rather have it merged in as is, I can revise it for you

3vcloud:
Actually looks like you've nailed it with your last commit, good job

careyourcurroundings:
> Actually looks like you've nailed it with your last commit, good job

Thanks! You've been very patient and great at explaining.

careyourcurroundings:
New feature, new PR:
https://github.com/careyourcurroundings/GWToolboxpp/pull/1

PR #2154, opened 2026-05-28
Fix pointer-based ImGui IDs that pollute ini files

### LINKED ISSUE #3: camera unlock
Originally reported by: **Anonymous**

---

Hey Has,
Is there a way to include the previous tb's camera zoom unlock in tb++? It was a very nice, scenic, often useful advantage. Thanks!

---
- Bitbucket: https://bitbucket.org/ggori/gwtoolbox-plusplus/issue/3

### LINKED ISSUE #2152: Layouts.ini and interface.ini files full of duplicate/obsolete info
Happened to notice that in both the Layouts.ini and interface.ini files, there were many listings (thousands) with windows position/size data for obsolete builds that had been deleted or altered. Most were listed as herobuild####, but no build name. One I noticed, had more than 60 listings with position/size data for the same build number. I only have 57 hero builds. There also seems to be a large amount of redundancy between the files. Perhaps they could be consolidated for greater efficiency.  BTW, with the update from 8.17 to 8.19, my interface.ini went from about 1200 lines to 8600 and Layouts.ini went from 5700 to 15000. Seems like there may be a problem there.

3vcloud:
Fixes #3 instances where runtime pointer values (memory addresses) were being used as ImGui widget/popup IDs. Since ASLR changes addresses each run, this generated new unique entries in interface.ini/Layouts.ini on every session, contributing to the bloat described in #2152.

Changes:
- `AudioSettings.cpp`: reorder buf setup before PushID, use string content
- `ImGuiAddons.cpp` SmallConfirmButton: remove dead id_buf line (never used)
- `ImGuiAddons.cpp` ConfirmButton: replace %p pointer with label-based popup ID
- `GWMarketWindow.cpp`: use item.name->c_str() instead of raw std::string*

Closes #2152

### DIFF
--- a/GWToolboxdll/ImGuiAddons.cpp
+++ b/GWToolboxdll/ImGuiAddons.cpp
@@ -347,7 +347,8 @@ namespace ImGui {
     bool ConfirmButton(const char* label, bool* confirm_bool, const char* confirm_content)
     {
         static char id_buf[128];
-        snprintf(id_buf, sizeof(id_buf), "%s##confirm_popup", label);
+        const ImGuiID h = ImHashStr(confirm_content, 0, ImHashStr(label));
+        snprintf(id_buf, sizeof(id_buf), "%s###confirm_popup_%u", label, h);
         if (Button(label)) {
             OpenPopup(id_buf);
         }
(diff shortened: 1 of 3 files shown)


3vcloud:
@claude how confident are we that the change to the confirm popup id don't interfere with other calls to this logic?

3vcloud:
@claude apply your suggestion

3vcloud:
@claude iirc the confirm content is optional so could default to the same string as something else that calls it - ensure anywhere calling this logic has suitable PushID's given

Issue #2150, opened 2026-05-28
custom circle markers scrambling with each restart


randomverbage:
All my custom circle markers, but not lines or polygons, are a mess. There's about 600 in my list. The sequence got completely scrambled with the update from 8.17 to 8.19. Notably, they're also getting re-scrambled with every restart.

3vcloud:
@claude 2 issues to sort here - first, why are they getting scrambled, and secondly, can anything be done easily to restore their original order?

randomverbage:
perhaps sorting by map ID

3vcloud:
> perhaps sorting by map ID

@claude this is a good shout; add an option to sort by map aswell as the A-Z implementation

PR #2156, opened 2026-05-28
fix: sort sections alphabetically in ToolboxIni to fix marker scrambling

3vcloud:
Fixes custom circle markers being scrambled on every restart (#2150).

The replacement of CSimpleIni with ToolboxIni changed section storage from std::multimap (alphabetical iteration order) to std::unordered_map (no guaranteed order). This caused GetAllSections() and SaveFile() to iterate sections in non-deterministic hash-bucket order, scrambling markers on every save/load cycle.

Restores the previous behaviour by sorting sections alphabetically in both SaveFile() and rebuildSectionCache().

### DIFF
--- a/GWToolboxdll/ToolboxIni.cpp
+++ b/GWToolboxdll/ToolboxIni.cpp
@@ -225,11 +225,21 @@ void ToolboxIni::LoadBuffer(std::string_view buf) {
 }
 
 SI_Error ToolboxIni::SaveFile(const std::filesystem::path& path, bool addUtf8BOM) const {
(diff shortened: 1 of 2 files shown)


3vcloud:
@claude can we at least amend the code to make custom marker sorting easier at a ui level?

3vcloud:
@claude your sorting buttons should be ConfirmButtons because its a pretty significant change for someone that has a lot of markers.

Also introduce the move up/move down type buttons similar to the ones that are used in hotkeys/herobuilds. Consider how this info will be saved to the ini

PR #2159, opened 2026-05-29
fix: crash when saving "hide when salvaging" list with rare materials or Nicholas items

### LINKED ISSUE #2141: App crash after adding 'items to hide when salvaging'
App crash when upon saving in setting panel after adding 'items to hide when salvaging' list, if the items are in the groups of items defined by the Salvage Rare Materials and/or Salvaging Nicholas Item checkboxes (checked or unchecked). Didn't try them all, but saw that pattern starting.

3vcloud:
Fixes #2141

GW item name_enc strings for certain items (rare materials, some Nicholas items) contain lone Unicode surrogates (e.g. LumpofCharcoal = L"\x22D1\xDE2A\xED03\x2625" where \xDE2A is a low surrogate in range 0xDC00-0xDFFF).

WideCharToMultiByte with WC_ERR_INVALID_CHARS rejects lone surrogates on Vista+, so WStringToString asserted and crashed when MapToIni tried to serialize block_from_being_salvaged keys after the user added one of those items via "Hide this when salvaging".

**Fix:**
- Switch wstring-keyed MapToIni to hex encoding (ArrayToIni) which is lossless for any wchar_t value
- IniToMap gains a StringToWString fallback for configs written before this change
- WStringToString gains a lossy substitution fallback instead of asserting, as a safety net

### DIFF
--- a/GWToolboxdll/Utils/GuiUtils.h
+++ b/GWToolboxdll/Utils/GuiUtils.h
@@ -37,14 +37,24 @@ namespace GuiUtils {
     // Reposition a rect within its container to make sure it isn't overflowing it.
     ImVec4& ClampRect(ImVec4& rect, const ImVec4& viewport);
 
+    // Takes a wstring and translates into a string of hex values, separated by spaces.
+    // Lossless: handles all wchar_t values including lone Unicode surrogates present in GW encoded strings.
+    bool ArrayToIni(const std::wstring& in, std::string* out);
+    bool ArrayToIni(const uint32_t* in, size_t len, std::string* out);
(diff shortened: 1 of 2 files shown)


3vcloud:
@claude none of this shit; just make sure anywhere that is trying to write encoded wstrings to the ini does so via ArrayToIni instead of WStringToString

3vcloud:
@claude would player names that have kanji or korean characters in them still be safe to write to the ini, or should those be amended too?

Maybe we should define a WStringToIni function signature to make it more obvious?

PR #2160, opened 2026-05-29
Fix Layout.ini bloat from window position accumulation

### LINKED ISSUE #2152: Layouts.ini and interface.ini files full of duplicate/obsolete info

3vcloud:
Fixes five root causes of Layout.ini growing unboundedly on every session (issue #2152):

1. `SaveUILayout()` now skips child windows, popups, tooltips, and other ephemeral ImGui widgets
2. Window keys now use only the stable ID after `###`/`##`, so renaming a teambuild no longer orphans old entries
3. The `[Windows]` section is cleared before each save, so deleted windows don't accumulate
4. `LoadUILayout()` pre-populate now correctly extracts window IDs from keys (fixed underscore parsing)
5. `TeamBuild::ui_id` is now persisted in herobuilds.ini and builds.ini so window positions survive reordering

Closes #2152

### DIFF
--- a/GWToolboxdll/Modules/ToolboxTheme.cpp
+++ b/GWToolboxdll/Modules/ToolboxTheme.cpp
@@ -14,6 +14,19 @@ constexpr auto WindowPositionsFilename = L"Layout.ini";
 constexpr auto IniSection = "Theme";
 
 namespace {
+    // Returns the stable save identifier for a window name.
+    // For "Display Name###ID" returns "ID" (the part after ###).
+    // For "Label##ID" returns "ID" (the part after ##).
+    // For plain names returns the full name.
+    const char* GetWindowSaveId(const char* name)
+    {
+        const char* triple = strstr(name, "###");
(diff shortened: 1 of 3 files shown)


3vcloud:
@claude if i ran toolbox, but didnt use the armory window for example, would it still remember where i put it even though i didnt trigger the window on this session? I suspect your change that nukes the ini section before rewriting would impact this

PR #2166, opened 2026-05-30
Add per-build Load buttons and attributes to detached teambuild view

### LINKED ISSUE #2164: Add Load button(s) to available skillbars when viewing a Teambuild
When clicking a teambuild from chat to show its window, we can't do much. For each build, if its a hero build and that hero is unlocked, then we can load it. If its a player build and we match the profession we can load it.

3vcloud:
Closes #2164

When viewing a teambuild from a chat link, each skillbar now shows:
- Build attributes (profession, attribute names and values)
- A Load button enabled when the hero is unlocked (hero builds) or the player's primary profession matches the build (player builds); disabled with a tooltip explaining why when conditions aren't met

### DIFF
--- a/GWToolboxdll/Utils/TeamBuild.cpp
+++ b/GWToolboxdll/Utils/TeamBuild.cpp
@@ -53,6 +53,7 @@ namespace {
 
     std::vector<PendingBuildLoad> pending_build_loads;
     std::queue<std::wstring>      send_queue;
+    std::string                   pending_reroll_build_code;
     clock_t send_timer   = 0;
     clock_t kickall_timer = 0;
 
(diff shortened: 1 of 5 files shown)


3vcloud:
@claude if its a player build, and the player has an AvailableChar that they could reroll to in order to load the build, then the "Load" button becomes "Reroll" - it'll call RerollWindow::Reroll (and let the reroll module decide who should be rerolled for the matching profession), and ensure that the bools for rejoining the party are true just in case we're in a team. Once the reroll process is finished (maybe expose RerollWindow::IsRerolling), load the build.

This gives need for the reroll window to have a new section called "Preferred Characters" - this allows a player to choose an AvailableChar that matches each of the 10 professions, e.g. if the reroll is asking for an elementalist, it'll check this list to see if it should look for a specific player before falling back to the normal logic.

This might need use to also expose RerollWindow::RerollToProfession so the reroll window can do this logic. Also consider refactoring the reroll chat command to use this new code as applicable

3vcloud:
@claude preferred profession chars should be unique per account - i think AccountMgr exposes the UUID. Preferred profs should be in the settings for the window, not the window itself. The new functions you've defined for the RerollWindow should be static, no need to get Instance().

Is this load/reroll logic also added to the detached draw of a player teambuild?

3vcloud:
move DrawDetachedTeambuild from the hero builds window code into the teambuild class

3vcloud:
@claude 
move DrawDetachedTeambuild from the hero builds window code into the teambuild class

3vcloud:
@claude the rerolling code thats in the update loop of hero builds window should be in Build::Update to detach this logic from the hero builds window

PR #2168, opened 2026-05-30
feat: add UI message callback timing to Performance window

3vcloud:
Implements # 2167.

Adds per-module UI message callback timing to the Performance window:

- `ToolboxModule::RegisterUIMessageCallback()` wraps each callback with QPC timing
- 44 call sites across 35 files updated to use the instrumented wrapper
- Performance window gains tabbed layout: "Modules" (existing) + "UI Messages" (new)
- UI Messages tab shows min/avg/max per-invocation timing and call count over 5s rolling window

### DIFF
--- a/GWToolboxdll/GWToolbox.cpp
+++ b/GWToolboxdll/GWToolbox.cpp
@@ -723,6 +723,11 @@ void GWToolbox::SetProfilingEnabled(bool enabled)
     profiling_enabled = enabled;
 }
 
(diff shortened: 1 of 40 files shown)


3vcloud:
@claude this is better, but it still doesn't tell me "this is the ui message id for this module that took a long time" - its just a collated summary per module.

3vcloud:
@claude is an unordered_map more performant here?

PR #2173, opened 2026-05-30
Skill Based Damage Tracking Via Temporal Inference

careyourcurroundings:
This is very much WiP. It will likely take a long time before it's ready. The code is a huge mess so I wouldn't look at it yet, but I thought it might be a good idea to get some feedback before I invest a lot of time into this.
The idea is to use the new max_hp cache to calculate damage, and to match the damage packets to skill or projectile events.
Outgoing damage, ie damage done by your party members is mostly working. Incoming damage, ie damage done to your party members is mostly not working. DoTs that have set ticks track very nicely (think firestorm), but variable delayed damage do not (think mistrust or spiteful spirit).

1/ Think my next step is to convert from matching skill->damage to damage->skill. I originally went skill->damage because I thought it would be more efficient to match one event with multiple damage. But I've since realized that there's a problem due to skills having a delay until they do damage. I also need to move the matching logic from draw to a damage packet callback. There's a few bugs I'm aware of. There's a lot of code that can be simplified, and all the code can be organized.

2/ I also thought about predicting damage and using a combination of time-based and damage-based matching. That would likely mean importing third party skill data and keeping a cache of enemy armor values, for each element. It would also mean a lot more work. Skills would need to be cross referenced to other party member's damage packets, ie splinter weapon's damage count for attacker not caster. Variable delay skills would track more accurately.

3/ Need some feedback on what features are actually useful.

4/ Projectile impact timing is currently predicted using a 1600 gwinch/second. It's accurate for wands and some bows, but can stray for fast spell projectiles. Not really sure how to fix that, especially for enemy projectiles. I also need to change projectile distance calc from distance to target to destination.
I'm not 100% sure that projectiles travel at a fixed distance per second.

5/ Dagger's dual attack's second hit is delayed 300ms from the first. AgentLiving's weapontype doesn't seem to work for predicting whether a target has a dagger equipped or not. I haven't tried NPCEquipment yet.

6/ Probably the hardest thing to fix would be long duration variable delay heal spells like PwK. Any ideas? Is damage matching the only way to go?

image
.

.

.

.

image

### DIFF
--- a/GWToolboxdll/Modules/ToolboxSettings.cpp
+++ b/GWToolboxdll/Modules/ToolboxSettings.cpp
@@ -63,6 +63,9 @@
 #include <Windows/InfoWindow.h>
 #include <Windows/MaterialsWindow.h>
 #include <Windows/NotePadWindow.h>
+#include <Windows/PartyStatisticsWindow.h>
+#include <Windows/EnemyStatisticsWindow.h>
+#include <Windows/TradeWindow.h>
 #include <Windows/ObjectiveTimerWindow.h>
 #include <Windows/ObserverExportWindow.h>
 #include <Windows/ObserverPartyWindow.h>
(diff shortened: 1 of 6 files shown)


3vcloud:
I don't have the drive to read through and comment on this - at the end of the day I don't think it's even nearly worth the effort needed to get even close to an accurate result. I'm just not that interested in damage by skill to go down this rabbit hole at all, but if it's something you're passionate about then good luck with it!

PR #2186, opened 2026-05-31
Highlight locked areas on the world map

JulienTant:
## Summary
Add an option to highlight locked areas when showing all areas on the world map.

it also centralizes repeated map display/exclusion checks in GW::Map helpers.

## Screenshot
Screenshot_20260531_114819


3vcloud:
I was expecting to find loads of extra code lines, but you've actually tidied up some of the map code while you were at it - nice job

PR #2194, opened 2026-06-01
fix: hotkey group trigger propagation and Map ID shared-state bugs

### LINKED ISSUE #2192: hotkey issues
@claude It appears that with the newest version, my hotkeys no longer automatically trigger when I enter an explorable area.
I could check "Trigger hotkey when entering explorable area" for the group settings, and it would apply for every hotkey in the group. That no longer works. I have to click it individually now.
Also... Map IDs is still messed up. Not only does it not work, but adding Map ID to ANY hotkey changes it for EVERY hotkey.

3vcloud:
Fixes two bugs reported in #2192:

- Group "Trigger hotkey when entering explorable area" setting now propagates to all child hotkeys by walking the group hierarchy in OnMapChanged()
- Adding a Map ID to one hotkey no longer affects all hotkeys (fixed ImGui triple-### ID bug and broken duplicate check using pointer address instead of value)
- Same ID fix applied to Character Names input

Closes #2192

### DIFF
--- a/GWToolboxdll/Windows/Hotkeys/TBHotkey.cpp
+++ b/GWToolboxdll/Windows/Hotkeys/TBHotkey.cpp
@@ -620,16 +620,15 @@ bool TBHotkey::DrawSettings()
             ImGui::PopID();
         }
 
-        static char map_id_input_buf[4];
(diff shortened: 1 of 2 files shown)


3vcloud:
@claude does the current hotkey code prevent a hotkey from being triggered twice? (it should)

PR #2209, opened 2026-06-04
perf: speed up AccountInventory load for players with many files

### LINKED ISSUE #2208: account inv module slowing down load
@claude the account inventory module's load logic is slow for players that have a lot of files. Assess the process in this module and see if we can improve it

3vcloud:
Fixes three performance issues in `LoadFromFiles` in `AccountInventoryWindow.cpp`:

1. Fix O(N²) entries accumulation - `GetAllSections` appended without clearing, causing O(N²) section processing for N files
2. Replace `CreateFileW`/`GetFileTime`/`CloseHandle` in `CheckIniDirty` with `file.last_write_time() `from the directory entry (already fetched for free)
3. Add `is_loaded` flag to skip re-processing unchanged files on repeated calls

Closes #2208

### DIFF
--- a/GWToolboxdll/Windows/AccountInventoryWindow.cpp
+++ b/GWToolboxdll/Windows/AccountInventoryWindow.cpp
@@ -1,6 +1,7 @@
 #include "stdafx.h"
 #include <atomic>
 #include <bit>
+#include <mutex>
 #include <GWCA/GameEntities/Agent.h>
 #include <GWCA/GameEntities/Map.h>
 #include <GWCA/Context/CharContext.h>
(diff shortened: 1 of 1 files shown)


3vcloud:
@claude use Resources::EnqueueWorkerTask to load separate files async. Only run 4 worker tasks at any time in this process, maybe batch them

3vcloud:
@claude the loading logic is too long, messy and unreadable. Create some helper functions to chunk it up e.g. IniToLoadResult etc. Don't bother with EnqueueMainTask; use some sort of mutex if you need to instead. The code that called EnqueueWorkerTask should then just use Sleep() for up to 15 seconds to wait to the enqueued stuff to finish

PR #2221, opened 2026-06-07
fix: hero build edits now persist and allow re-adding deleted slots

### LINKED ISSUE #2218: Hero Builds Editing
Editing hero builds after being created does not save changes after saving or closing and reopening guildwars.

Builds can be deleted from herobuilds but cannot be added - there is no way to add back a player build or hero build once it has been deleted.

3vcloud:
Fixes #2218

**Bug 1 — Changes not saved:** In `DrawHeroBuildsContent` and `DrawPlayerBuildsContent`, several edit operations never set `builds_changed = true`: per-build name edits, per-build code edits, behavior toggle, disabled-skills toggle, and pcon toggling. Since `SaveToFile()` is guarded on `builds_changed`, all these edits silently failed to persist.

**Bug 2 — No "Add" button:** `DrawHeroBuildsContent` only offered Delete in the per-row dropdown. Added "Add Hero Slot" and "Add Player Slot" buttons at the bottom of the edit window.

### DIFF
--- a/GWToolboxdll/Utils/TeamBuild.cpp
+++ b/GWToolboxdll/Utils/TeamBuild.cpp
@@ -595,7 +595,7 @@ void TeamBuild::Load() const
 // ------------------------------------------------------------
 // Player-builds layout (BuildsWindow style)
 // ------------------------------------------------------------
-void TeamBuild::DrawPlayerBuildsContent(bool& builds_changed, bool editable)
+void TeamBuild::DrawPlayerBuildsContent(bool& builds_modified, bool editable)
 {
     const float font_scale = ImGui::FontScale();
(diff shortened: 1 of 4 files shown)


3vcloud:
@claude builds_changed is actually a flag as to whether the underlying vector of teambuilds has changed - its just to know whether we should drop out of the loop because the iterator is now invalid. Rename this variable accordingly, and revert your changes that assigned it to true in cases where the vector would not have been invalidated.

3vcloud:
@claude there are still some cases where the flag is set to true when the vector being looped would not have been modified.

3vcloud:
@claude consider what would happen when the Group field is edited.

PR #2230, opened 2026-06-11
turn all setting files into .json files in configs/default folder def…

henderkes:
…ault, keep migration path from ini files until 9.0

image

### DIFF
--- a/GWToolboxdll/Modules/ChatCommands.cpp
+++ b/GWToolboxdll/Modules/ChatCommands.cpp
@@ -704,6 +704,8 @@ namespace {
 
     std::vector<CmdAlias*> cmd_aliases;
 
+    ChatCommands::Settings settings;
+
     void sort_cmd_aliases()
     {
         std::ranges::stable_sort(cmd_aliases, [](const auto* a, const auto* b) {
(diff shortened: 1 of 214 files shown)


GWToolboxdll/Modules/ChatCommands.cpp
```diff
@@ -1134,57 +1024,286 @@ namespace {
...
+            case SettingsRegistry::Type::Int:
+            case SettingsRegistry::Type::Uint:
+            case SettingsRegistry::Type::Float: return "<number>";
+            case SettingsRegistry::Type::Color: return "<0xAARRGGBB>";
+            case SettingsRegistry::Type::Float2: return "<x> <y>";
+            default: return "<text>";
+        }
+    }
+
+    std::string SettingValueToString(const SettingsRegistry::Entry& entry)
```
3vcloud, line 1052: a graceful way of handling encoded wstrings would be nice - you cant just convert WStringToString because it uses naughty characters, so atm we to ArrayToIni

henderkes: we still use ArrayToIni and IniToArray because of backwards compatibility. throwing those out for the wstring overload can happen in 9.0 (hopefully still a bit away to give users time to migrate)

PR #2245, opened 2026-06-12
Disable hero build load button when party is full

### LINKED ISSUE #2242: Hero Builds are not loaded
Heroes can be added to party using TB but the function to load builds doesn’t work.

3vcloud:
Amend the individual hero build load button in `DrawHeroBuildsContent`: disabled with tooltip "No space in the party to load the hero" when the party is full and the hero is not already in the party.

Closes #2242


3vcloud:
@claude Include Map.h

PR #2248, opened 2026-06-14
Add multi-map pathfinding

544449:
Cross-map portal routing via a shipped portal_connections.json plus an on-demand .dat loader that builds pathing data for maps the player isn't in, so a full route can be computed across several maps. The world-map "Place Marker" action draws the full cross-map route on the world map.

Credits: trapezoid/BSP pathfinding from OpenTyria (https://github.com/ldufr/OpenTyria); GW .dat parsing reference from GuildWarsMapBrowser (https://github.com/Jonathan-Greve/GuildWarsMapBrowser).


3vcloud:
Absolutely love this, will take a look this evening thanks

3vcloud:
@claude this has been merged into dev, but can we reduce the footprint of the map constant data in a new branch by leveraging GW::Constants::MapID, GW::Map::GetMapInfo and Resources::GetMapName to avoid the need for hard coded strings and the is_pvp flag?

PR #2258, opened 2026-06-17
feat: add Backup module for settings archive and restore

### LINKED ISSUE #2257: Backup module
@claude backing up gwtoolbox settings is currently a manual process. Create a module that ultimately creates an archive of the gwtoolbox settings folder with settings to tell it what types of files (e.g. just text files or images and mp3s aswell) as a series of checkboxes. Archives should be named appropriately with the current date/time.

Also provide the user with a button to browse for an already created archive file to restore using the file browser functionality already in the code. When the archive is chosen and is valid, ask the user questions regarding overwriting etc, and also if they'd like to first backup the current files before restoring the chosen archive.

Whenever toolbox is updated, it needs to first call this module to create a text file backup, labelled appropriately just in case the user wants to roll back by restoring the automatically created archive.

Also consider how the launcher updates toolbox - it needs to do similar behaviour to the above when updating.

Be careful not to add dmp or zip files to the archive - perhaps only allow certain file extensions to be detected.

By default only text files should be backed up.

3vcloud:
Implements a new Backup module that creates and restores ZIP archives of the GWToolbox settings folder.

**Features:**
- File type filtering checkboxes (text on by default, images/audio optional)
- Self-contained ZIP writer/reader (no new dependencies)
- Restore dialog with "Backup & Restore"/"Restore Now"/"Cancel" options
- Auto-backup before each toolbox update via Updater::DoUpdate
- Auto-backups in `<settings_folder>/backups/`

Closes #2257

### DIFF
--- a/GWToolboxdll/Modules/BackupModule.cpp
+++ b/GWToolboxdll/Modules/BackupModule.cpp
@@ -7,38 +7,12 @@
 #include <Utils/TextUtils.h>
 
 #include "BackupModule.h"
+#include "CodeOptimiserModule.h"
 
 // ============================================================
 // ZIP (STORE) writer/reader — no external dependencies
 // ============================================================
 namespace {
-
-    // --- CRC-32 -----------------------------------------------------------
-
-    static uint32_t crc32_table[256];
-    static std::once_flag crc32_once;
-
-    static void ensure_crc32()
-    {
-        std::call_once(crc32_once, [] {
-            for (uint32_t n = 0; n < 256; n++) {
-                uint32_t c = n;
-                for (int k = 0; k < 8; k++)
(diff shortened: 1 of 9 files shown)


3vcloud:
@claude a lot of this code looks like pretty boilerplate stuff - take a look across the codebase to see if we can recycle any code, or if we can't but similar logic is used elsewhere, consider consolidating any shared functions as utility functions

3vcloud:
@claude do better - we literally have a module that overrides gw's crc32 algorithm, can't you amend that module to expose a static function to do the hash for you?

Further to this, can the Resources version ask the aforementioned function to do the heavy lifting and reduce some code there too?

PR #2261, opened 2026-06-17
Add DPS To Party Damage Widget

careyourcurroundings:
Calculates battle segment duration.
Segment start is first damage packet time.
Segment end is last damage packet time, which is defined when 5 seconds have elapsed with no damage packets.
DPS is calculated using total damage divided by accumulated battle segment duration.

Should work in sc splits. If battles take place out of compass range, you won't get those packets, and won't accumulate battle time.

image


careyourcurroundings:
Added condition DPS:
image

3vcloud:
condition dps sources can't reliably be tracked

careyourcurroundings:
> condition dps sources can't reliably be tracked


(1)
Sources aren't tracked. It tracks total damage done by condition regardless of source, assuming that -10 pips isn't hit.

Here's the method:
When packet is sent to client telling it to apply a condition to an enemy, track that enemy.
When packet is sent to client telling it to remove a condition, that enemy died, or enemy despawned/out of range, remove the tracking, and calculate total damage done by that condition by multiplying degen by time affected by condition.

(2)
I'm happy to edit the tracking if you can think of a better method. Or remove it entirely, but a good amount of people have expressed that they want dot tracking in toolbox.

(3)
I could also cap the degen at -10 pips per enemy, and split the damage between the conditions. I.e. if enemy has poison+disease+burning for -15 pips, I would do 4/15, 4/15, 7/15. It seems a bit complicated and overkill to do that though.

(4)
If you agree with this, I could also track hex degens. I don't think there are that many hex degens in this game that manually inputting the pips would be a lot of work. Deep Wound could also be tracked (20% or 100 damage if enemy dies with deep wound). I just don't want to do extra work if you think this is unviable.

3vcloud:
I've taken a look at this today, but still having a bit of trouble understanding the condition damage thing? What is it supposed to represent?

I tried it out in minister chos today and it seemed to get stuck on bleeding at 4/s - The bleeding wasn't done by me, I thought it might be a current target thing but my target died ages ago, but isnt bleeding/poison/disease etc a fixed amount of degen anyway, so we'd already know the degen?

3vcloud:
> I've taken a look at this today, but still having a bit of trouble understanding the condition damage thing? What is it supposed to represent?
> 
> I tried it out in minister chos today and it seemed to get stuck on bleeding at 4/s - The bleeding wasn't done by me, I thought it might be a current target thing but my target died ages ago, but isnt bleeding/poison/disease etc a fixed amount of degen anyway, so we'd already know the degen?

I think this is something to do with the combat timer - dps doesn't clear either

careyourcurroundings:
> I've taken a look at this today, but still having a bit of trouble understanding the condition damage thing? What is it supposed to represent?
> 
> I tried it out in minister chos today and it seemed to get stuck on bleeding at 4/s - The bleeding wasn't done by me, I thought it might be a current target thing but my target died ages ago, but isnt bleeding/poison/disease etc a fixed amount of degen anyway, so we'd already know the degen?

Afflicted Necromancers in minister cho’s self-bleed with blood drinker. There’s no easy way to tell who is responsible for the bleed. EotN dungeon traps afflicting enemies with conditions will run into same problem.

4/s bleed means that while you are in combat, 0.67 enemies have bleed condition on them.

3vcloud:
> 4/s bleed means that while you are in combat, 0.67 enemies have bleed condition on them.

I'm still a little confused about this; @claude can you take a look at the logic around the condition tracking in this code and try to explain what it represents?

careyourcurroundings:
Damn, claude makes mistakes but it is really good at telling you where potential bugs are. I'm going add asking claude to check for bugs to my workflow.
Condition tracking is through GenericValue packet. I think add_effect is for your party's enchants and hexes.
1. Definitely a bug. I accidentally deleted the callback registration. But it doesn't add phantom damage like claude claims b/c damage never gets finalized. It leads to the accumulation of trackers, which eventually fill up the 64 sized array.
2. Definitely a bug. Bug is due to wrong scope in ci definition, which like claude said, leads to removal of all conditions when one expires, undercounting the damage.
3. Not a bug. DPS shouldn't be cleared between battles.

Fixed and refactored some code.

3vcloud:
thanks, I'll test again this evening

PR #2299, opened 2026-06-21
Remove Termius section from README

henderkes:
Removed Termius logo and description from README.


henderkes:
@claude any other place it's still mentioned? Please remove.

PR #2315, opened 2026-06-22
Fix Saving Disabled Hero Skills

careyourcurroundings:
Bug was way simpler than I thought


3vcloud:
@claude amend the release notes to add this fix

PR #2330, opened 2026-06-26
Add Playmate companion plugin

zalex230:
First fork and release


3vcloud:
Plugins aren't maintained, and are just for reference, but this one is automating gameplay on another account so would be a no to integrate

PR #2366, opened 2026-07-03
feat(worldmap): draw multimap path at any zoom, incl. animation

544449:
The world-map path/line overlay inlined its own world->screen transform and dropped the world_map_scale factor, so it only aligned when fully zoomed in. Route the line and navmesh blocks through the shared CalculateViewportPos helper instead, and make that helper animation-correct: derive a per-frame pixels-per-world-coord scale from the live visible rect (top_left..bottom_right, which the client re-derives every frame from the animating zoom) and the world-map frame's on-screen size, falling back to the rest-state factor until the client populates the rect.

With projection now correct at every zoom level, drop the five "Map is animating" early-returns so quest/boss/area/portal overlays track the zoom animation too, and scale the boss capture icon smoothly (std::lerp 16->32) instead of snapping between two sizes.

### DIFF
--- a/GWToolboxdll/Widgets/WorldMapWidget.cpp
+++ b/GWToolboxdll/Widgets/WorldMapWidget.cpp
@@ -153,6 +153,7 @@ namespace {
     GW::Vec2f viewport_offset;
     GW::Vec2f ui_scale;
     float world_map_scale = 1.f;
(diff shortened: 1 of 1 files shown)


3vcloud:
A big change for a small amount of edits, very nice thanks

PR #2379, opened 2026-07-05
Pathfinding: portal-graph multimap router

544449:
Rewrite cross-map routing. Only use portal connections from portal_connections.json, replaces the bounds-overlap fallback.
Remove the deprecated OpenTyria trapezoid pathfinder.
Remove RunAStarOnMap's offset-nudge retries.
QuestModule: cross-map goal handling + per-path Update() throttle (a shared static was starving other quest paths).
World-map line viewport cull.
portal_connections.json 401->408 connections.
Quiet expected probe/route logs.


544449:
maybe wait with pulling this before other crashing issues are resolved.

PR #2404, opened 2026-07-12
perf(account-inventory): load inventory files off the game thread

3vcloud:
## Why loading in is slow with Account Inventory enabled

`LoadFromFiles()` already dispatched the JSON parse to worker threads — but then **busy-waited on them on the game thread**:

```cpp
const clock_t load_start = TIMER_INIT();
while (state->tasks_remaining > 0 && TIMER_DIFF(load_start) < 15000)
    Sleep(10);
```

So the async split bought nothing. This runs on:
- **module toggle** (`LoadSettings` → `LoadFromFiles(true)`)
- **every character switch / logout** (`kLogout` → `LoadFromFiles(true)`)
- toolbox init and Delete All (`LoadFromFiles(false)`)

i.e. exactly the "load in" moments. The game is frozen until every account file is read.

### Why it's ~10s even for 23 small files (1.8 MB) on a 20-thread pool

It isn't the parse cost. The game thread is held hostage to the shared worker pool:

- the parse is split into only **`min(4, files)` batches**, so at most 4 of the 20 workers are ever used;
- those tasks land in the **single FIFO queue shared with every other Toolbox background job** (item-icon DAT decodes, web fetches, …), behind whatever is already queued at load-in;
- idle workers only pick up new work on a **100 ms poll** (no wakeup signal in `Resources::WorkerThread`).

So with any backlog the `Sleep(10)` loop just spins on the game thread up to the 15 s cap. The pool size is irrelevant when the game thread is the one waiting.

## The change

Make the foreign-only reload (the map/character-change path) **fire-and-forget**:

- the worker fills a `shared_ptr<PendingLoad>` (`results` + a `done` flag);
- `Update()` calls `DrainPendingLoad()` and applies the results once `done`, without blocking;
- the **full** load (Initialize / Delete All), whose results the caller immediately builds on, still completes synchronously via `FlushPendingLoad()` so item ordering (disk baseline → live overwrite) is preserved.

### The mutex

Per review request, a `load_mutex` is **held by the worker for the whole duration of its file reads**. `Terminate()` and full-file saves (`SaveToFiles(true)` / Delete All) acquire it via `FlushPendingLoad()`, so the module can't be torn down or have its files rewritten/deleted while a load is still reading them. The worker sets `done` before releasing the mutex, so a released mutex implies a finished load; if the task never got scheduled before the timeout, `done` stays false and it's simply applied by a later `Update()`. The `PendingLoad` is `shared_ptr`-owned, so a superseded or post-`Terminate` worker finishes harmlessly into an orphan.

## Notes / trade-offs
- After a map/character change there's now a brief window (until the background load applies) where **other accounts'** items aren't shown yet; the current account is always present. This replaces a multi-second freeze with a barely-visible refresh.
- The deeper pool issues (100 ms idle poll, 4-batch cap, one shared FIFO) are untouched here — they now only affect how fast the *background* load finishes, not the game thread. Happy to follow up with a condition-variable wakeup for the pool if wanted.

### DIFF
--- a/GWToolboxdll/Windows/AccountInventoryWindow.cpp
+++ b/GWToolboxdll/Windows/AccountInventoryWindow.cpp
@@ -1010,19 +1010,12 @@ namespace {
         bool success = false;
     };
 
-    // One dispatched load. A worker thread fills `results` and sets `done`; the main
-    // thread applies them (in Update, or a blocking flush) once done. Held behind a
-    // shared_ptr so an in-flight worker never dangles if the module is torn down or a
-    // newer load supersedes this one mid-parse.
+    // Dispatched load: a worker fills results + sets done, the main thread applies it once done.
     struct PendingLoad {
         std::vector<LoadResult> results;
-        std::atomic<bool> done{false};
+        std::atomic<bool> done{false}; // shared_ptr-owned, so a superseded/torn-down worker orphans safely
     };
-    // Serialises file access only: a worker holds it around each file read, SyncAccountFile
-    // around each write/delete, so a save can't rewrite/delete a file mid-read. Parsing and
-    // JSON serialisation run outside it. (Completion is signalled by PendingLoad::done, not
-    // this mutex.)
-    std::mutex load_mutex;
+    std::mutex load_mutex; // guards file access only (worker read vs SyncAccountFile write); done signals completion
     std::shared_ptr<PendingLoad> active_load;
 
     // Encode the wide game string as fixed-width 4-hex code units with no separators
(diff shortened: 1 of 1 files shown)


3vcloud:
Follow-up (34f0211f): the actual pool flood was `ApplyItemJson`/`AddItem` calling `Resources::GetItemImage` per item at load time — one DAT decode queued per stored item, most never shown. Moved the fetch into `Draw` so only the icon actually being rendered decodes; the returned pointer is stable and cached on the `Item` on first draw, so it's once-per-shown-item instead of once-per-stored-item. Dyes are stashed at add time so the deferred fetch still gets the tinted icon. In detailed view or with the window closed, nothing decodes at all now.

Issue #2408, opened 2026-07-12
Skill Range Rings: large rings cause severe FPS drops


3vcloud:
## Summary

The **Skill Range Rings** module (disabled by default) causes severe frame-rate drops that scale with the radius of the drawn ring. Multiple users report dropping into single-digit FPS while hovering skills with large ranges.

Reported in #gwtoolbox-support by aaron and Llancarfan. Loot Beacon and Danger Rings are unaffected.

## Root cause

The rings are re-draped onto the terrain **every frame** in `SkillRangeRingsModule::DrawInWorld`. For each ring, `EmitBand()` samples the surface height per vertex via `TerrainDrape::SurfaceZ`, which in turn loops over **every pathing plane** calling the (expensive) `GW::Map::QueryAltitude`.

The vertex count scales with radius: `RingSegments()` clamps at up to **512 segments** for a large ring, and each segment samples both an inner and outer vertex. So per large ring, per frame:

```
512 segments × 2 vertices × n_planes  →  up to ~1000 × n_planes QueryAltitude calls
```

On maps with several pathing planes that's thousands of terrain-altitude queries every single frame — hence the FPS collapse, and why it gets worse the bigger the ring is.

Relevant code:
- `GWToolboxdll/Modules/SkillRangeRingsModule.cpp` — `DrawInWorld` / `EmitBand` (per-frame re-drape, per-vertex `SurfaceZ`)
- `GWToolboxdll/Utils/TerrainDrape.cpp` — `SurfaceZ` loops all pathing planes per call

## Possible directions

- Cache the draped ring geometry and only rebuild when the anchor position moves beyond a threshold, rather than every frame (the ring specs are already cached; the vertex drape is not).
- Reduce/cap the effective sampling density for large rings, or increase `kSampleSpacing`.
- Avoid redundant per-plane `QueryAltitude` scans (cache per (x,y) or reduce plane iteration).
- Reuse the inner/outer vertex buffers instead of allocating two `std::vector`s per `EmitBand` call.

## Notes

The module is disabled by default, partly for this reason.

3vcloud:
@claude link 2402 to this issue

PR #2414, opened 2026-07-12
perf(terrain): compute draped altitude natively instead of per-frame QueryAltitude

### LINKED ISSUE #2409: Investigate computing terrain altitude natively instead of GW::Map::QueryAltitude
## Motivation

Several in-world overlays drape geometry onto the terrain by sampling the surface height per vertex. Today this goes through `GW::Map::QueryAltitude` (`TerrainDrape::QueryAltAt` → `GW::Map::QueryAltitude`), which is a game-side call we invoke potentially thousands of times per frame. This is the dominant cost behind the Skill Range Rings FPS collapse (#2408): a large ring is ~1000 vertices × n_planes `QueryAltitude` calls **every frame**.

It would be worth investigating whether Toolbox can compute the altitude for an arbitrary (x, y[, zplane]) point itself, from data we already have mapped, avoiding the repeated dependence on the GW implementation.

## What we already have

GWCA exposes the full pathing map per zplane (`GW::Map::GetPathingMap()` → `PathingMapArray`), including:
- `PathingTrapezoid` — the 2D trapezoids (`XTL/XTR/YT`, `XBL/XBR/YB`) that tile each plane, plus adjacency and the XNode/YNode/SinkNode search tree (`root_node`) used for point location.
- `PathingMap.dat_vectors` and the node tree, which is exactly what a point-in-trapezoid lookup would traverse.

See `Dependencies/GWCA/include/GWCA/GameEntities/Pathing.h`.

## Open questions to investigate

1. **Where does the altitude value actually come from?** The `PathingTrapezoid` struct is purely 2D (no z). `QueryAltitude` must combine the trapezoid location with a terrain height source (heightmap / plane equation / interpolation across trapezoid corners). We need to find where that z data lives and whether it's reachable from Toolbox.
2. **Point location:** can we reuse the exposed node search tree (`root_node` + X/Y nodes → `SinkNode.trapezoid`) to locate the containing trapezoid cheaply, rather than scanning `trapezoids`?
3. **Interpolation:** once the trapezoid (and its height data) is known, how does GW interpolate altitude within it? Match that so our values agree with the game's.
4. **Correctness vs. cost:** even a native implementation is per-point work; the bigger win may be caching/decimation (see #2408). This issue is specifically about removing the hard dependency on `GW::Map::QueryAltitude` and enabling cheaper batched sampling.

## Consumers that would benefit

- `GWToolboxdll/Utils/TerrainDrape.cpp` (`SurfaceZ`, `DrapeZ`, `QueryAltAt`) — the single choke point; all draped overlays go through here.
- Skill Range Rings (#2408), and any future in-world draped geometry.

Related: #2408

henderkes:
Closes #2409. Also addresses the FPS collapse in #2408.

## Problem

Draped in-world overlays (skill range rings, loot beacons, danger rings, quest path, rivers, weather) sample the surface through `GW::Map::QueryAltitude` **once per plane per vertex, every frame**. That call swaps the global map context and walks the terrain heightfield, so a large ring (~1000 verts × N planes) collapses the frame rate.

## Change

**Read the terrain heightfield directly.** `TerrainDrape::NativeTerrainZ(x, y)` resolves the exact surface z at a point straight from the mapped terrain object (`MapContext+0x84`) — a 32×32-float-chunk grid, 96-gwinch tiles, Y-flipped, split on the B–C diagonal. It's a pure memory read: no game call, no context swap, thread-safe.

- `QueryAltAt` now routes **plane 0** (the common case) through `NativeTerrainZ`, so every draped consumer benefits with no change at the call sites. Non-zero planes still go through `QueryAltitude` for the prop-mesh raycast.
- `SurfaceZ` / `DrapeZ` no longer scan all planes. They take the native plane-0 surface as the baseline and query only the non-zero planes whose **pathing trapezoid actually contains (x, y)**, found via a point-location DAG walk — directly answering the issue's "query only the planes that exist" ask.

The terrain struct layout and query math were reverse-engineered from the client; `GwTerrain` is guarded by `offsetof` static_asserts.

## Verification (in-game, harness, map 474, 42 planes)

- **Correctness:** native z matches the client's own plane-0 value; the residual (avg ~2 gwinches) is one-signed and slope-correlated — the game's radius-5 disk-max versus point-exact interpolation, not a decode error.
- **Performance:** direct A/B on identical points, measured as an in-process CPU-timing loop (independent of window focus / frame pacing) → **~15× faster** per `SurfaceZ` call (21.5 → 1.46 µs).
- **Per-frame impact:** the skill-range rings rebuild every frame. A Sandstorm cast (312 + 1050-radius rings) is ~690 `SurfaceZ` calls/frame → **old ≈ 14.8 ms/frame** (a hard ~67 fps ceiling from ring sampling alone; the #2408 collapse) vs **new ≈ 1.0 ms/frame**. The largest allowed ring (512-segment cap, ~1024 samples) is old ≈ 22 ms vs new ≈ 1.5 ms.

Adds Debug-only harness verbs (`nativez`, `drapebench`, `fpsprobe`) used to produce the numbers above; these compile out of shipped builds.

## Behavior note

Plane-0 draping is now point-exact rather than radius-5-blurred like the client. This hugs the terrain more precisely — better for overlays — but is a deliberate behavior change worth flagging.

https://claude.ai/code/session_013qkJXMrytQ75KYP4o2yZgX

### DIFF
--- a/GWToolboxdll/Modules/TestHarness.cpp
+++ b/GWToolboxdll/Modules/TestHarness.cpp
@@ -3,6 +3,7 @@
 #include <chrono>
 #include <cmath>
 #include <filesystem>
+#include <set>
 #include <sstream>
 #include <string>
 #include <system_error>
@@ -14,6 +15,7 @@
 #include <GWCA/GameEntities/Skill.h>
 #include <GWCA/GameEntities/Quest.h>
 #include <GWCA/Context/CharContext.h>
+#include <GWCA/Context/MapContext.h>
 #include <GWCA/Context/PreGameContext.h>
 #include <GWCA/Constants/Constants.h>
 #include <GWCA/Managers/AgentMgr.h>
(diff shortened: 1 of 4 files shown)


henderkes:
### Equivalence check: pruned SurfaceZ vs. querying all planes

Verified over **1500 random positions** on the current map (474, 42 planes) via the new `drapeverify` harness verb. To separate the two independent changes, it compares:

- **plane selection** — the pruned query (only trapezoid-containing planes) vs. querying *all* planes, both using the game's `QueryAltitude` values so plane choice is the only variable
- **total** — the shipped `SurfaceZ` (native plane-0 read + pruning) vs. the old all-planes result

```
drapeverify: n=1500 both=1500 nonzero_plane_hits=38 | PRUNE mismatches=0 max=0.00 | total new-vs-old avg=6.59 max=128.38
```

- **Plane selection is identical: 0 mismatches / 1500, max 0.00** — including **38 points whose highest surface came from a non-zero bridge/prop plane**, every one selected the same plane the all-planes query did. So "only query the planes that exist at (x,y)" is provably equivalent to "query all planes, take the highest."
- The remaining `total` delta is purely the plane-0 sampling method: my native point-interpolation vs. the client's radius-5 disk-max. Avg ~2 gwinches (narrow) to ~7 (wide radius).
- The `max 128` outliers are all near-vertical cliffs, confirmed by sampling ±8 gwinches around each — e.g. at `(1860,-17161)` neighbors span −814 to −1019, a ~200-gwinch drop, so the game's 5-gwinch disk legitimately grabs a higher point than the exact vertex. On flat terrain the delta is ~0, which rules out an interpolation bug.

Takeaway: the plane-pruning is exact; the only intended value change is point-exact draping instead of radius-5-blurred.

PR #2435, opened 2026-07-24
Game Settings: +name color for Friends, Guildies, & Player's (In My Party)

Ryzzer:
Friends takes priority, then Guild, then Party:

Screenshot 2026-07-24 142754


3vcloud:
@claude this needs some sort of caching mechanism per map load; name tags are generated every time they're hovered. Keep a std::map<std::wstring, color> or something that would remember the colour without having to dive into the friend list or the guild member list etc every time. Clears on map change

PR #2438, opened 2026-07-26
fix(minimap): make Custom Agents size edits apply to other players

3vcloud:
## Summary
- Since #4dba954f unified agent colors/sizes into the Custom Agents list, real players (other party members) never reach `GetCustomAgentsToDraw` (there's an explicit "we don't support custom agents for players" note in `Render()`), so they always fell through to `GetSize()`'s allegiance switch.
- That switch read `size_ally`/`size_neutral`/`size_ally_spirit`/`size_ally_npc`/`size_minion` directly - fields that are now only ever written once, at migration/seed time. Editing the seeded "Neutral"/"Ally"/"Ally (NPC)"/"Ally (Spirit/Pet)"/"Ally (Minion)" rows in the Custom Agents list only updates the row's own copy, never that legacy field.
- Net effect: no UI control could change other players' (or dead/neutral NPC fallback) minimap size anymore - reported in Discord as "other players stuck at 100 no matter what I set".

## Fix
Added `GetSeededDefaultSize()`, which looks up the current size on the matching seeded default row (falling back to the legacy field only if that row is missing/inactive/disabled), and use it in `GetSize()`'s switch fallback for the 5 migrated allegiances. This only touches size resolution - color/shape for players are unchanged.

## Test plan
- [ ] In-game: set "Ally" size in Custom Agents settings to something other than 100, confirm other party members/players on the minimap resize accordingly.
- [ ] Confirm "Restore Defaults" still resets both the legacy field and the seeded row back to 100.
- [ ] Confirm NPC allies/neutrals/spirits/minions still size correctly (unchanged behavior, still matched via `GetCustomAgentsToDraw`).

### DIFF
--- a/GWToolboxdll/Widgets/Minimap/AgentRenderer.cpp
+++ b/GWToolboxdll/Widgets/Minimap/AgentRenderer.cpp
@@ -1249,6 +1249,20 @@ Color AgentRenderer::GetColor(const GW::Agent* agent, const CustomAgent* ca) con
     return IM_COL32(0, 0, 0, 0);
 }
 
+float AgentRenderer::GetSeededDefaultSize(const GW::Constants::Allegiance allegiance, const float fallback) const
+{
+    const auto it = custom_agents_by_allegiance.find(static_cast<int>(allegiance));
+    if (it == custom_agents_by_allegiance.end()) {
+        return fallback;
+    }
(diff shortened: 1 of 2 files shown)


3vcloud:
@claude add to release notes for this

PR #2458, opened 2026-08-07
fix: bound agent name copies in party window module

AlexMacocian:
RemoveAllyActual and AddAllyActual copied GW::Agents::GetAgentEncName() into a wchar_t[8] with an unbounded wcscpy. Encoded agent names routinely exceed 8 wchars, so this smashed the stack; GetAgentEncName can also return nullptr, which wcscpy would have dereferenced.

Size the buffers to the AgentName packet field they are eventually written back through, copy with bounds, and null-check the source. SetAgentName is hardened the same way so an over-long caller string truncates instead of overflowing packet.name_enc.

Reachable with default settings (add_npcs_to_party_window is on) in any explorable area. Found while investigating unrelated crash dumps; not known to be the cause of any reported crash.

### DIFF
--- a/GWToolboxdll/Modules/PartyWindowModule.cpp
+++ b/GWToolboxdll/Modules/PartyWindowModule.cpp
@@ -100,6 +100,9 @@ namespace {
     GW::HookEntry Summon_AgentAdd_Entry;
     GW::HookEntry Summon_GameThreadCallback_Entry;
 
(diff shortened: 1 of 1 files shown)


3vcloud:
This is wrong, enc agent names for allies never exceed 8 characters

AlexMacocian:
Sorry, I'll pay more attention at what claude is doing

PR #2460, opened 2026-08-08
fix(GameSettings): restore Enter Mission button on party leader change

3vcloud:
## The bug

If party leadership changes in an outpost, the new leader's **Enter Mission** button is gone until they toggle the game's "legacy Enter Mission button" option off and on.

## Root cause

Two UI messages are involved in the party window's mission button:

- **`kDisableEnterMissionBtn` (0x1000012a)** - the handler at `0x0056a110` looks up child frame 4 and calls `SetFrameVisible(child, ...)` from client state. This is the **only** place the button's visibility is set.
- **`kPartyLeaderChanged` (0x10000122)** - routes to `0x0056a780`, which creates/destroys the button frame and sets its *enabled* state from "am I leader" (`0x008567e0`, bit 7 of the party context flags). It never touches visibility.

A change of leadership broadcasts 0x10000122 only, so the visibility check never re-runs and the button stays hidden. Toggling the preference happens to work because it rebuilds the whole party window.

## The fix

Hook `kPartyLeaderChanged` at post altitude and send `kDisableEnterMissionBtn` ourselves, so the game runs its own visibility check. No preference twiddling, no party state checks on our side - the client recomputes everything.

`wparam` is 0: every subscriber of 0x1000012a (`0x0056a110`, `0x0056c0a0`, `0x00586a90`, `0x00565410`, `0x00565990`) ignores it and recomputes from client state, so the value can't push the button into a wrong state.

Also names 0x10000122 as `kPartyLeaderChanged` in the vendored GWCA `UIMessage` enum (was `kMessage_0x10000122`, unreferenced) - worth mirroring upstream in GWCA.

New checkbox under Game Settings -> Party, on by default, plus a docs entry in `settings.mdx`.

## Testing

Not compiled or run - no Windows/MSVC toolchain here, and reproducing needs two accounts in a party. Please have someone verify in-game before merging.

### DIFF
--- a/GWToolboxdll/Modules/GameSettings.cpp
+++ b/GWToolboxdll/Modules/GameSettings.cpp
@@ -941,6 +941,24 @@ namespace {
         }
     }
 
+    // Last party leader state the server told us about, so we can spot a change of leadership
+    bool was_party_leader = true;
+
+    void OnPartyLeaderChanged(GW::HookStatus*, const GW::Packet::StoC::PlayerIsPartyLeader* packet)
+    {
+        const auto is_leader = packet->is_leader != 0;
+        const auto became_leader = is_leader && !was_party_leader;
+        was_party_leader = is_leader;
+        if (!became_leader || !settings.fix_legacy_enter_mission_button) return;
+        if (!GW::UI::GetPreference(GW::UI::FlagPreference::LegacyStartMissionButton)) return;
+        if (GW::Map::GetInstanceType() != GW::Constants::InstanceType::Outpost) return;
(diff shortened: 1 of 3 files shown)


3vcloud:
I was joking when we said we could path this bug - don't fix their shit for them

3vcloud:
In fact, @claude this could literally be a 1 liner to forcibly send the update ui message manually, none of this preference toggle shit

3vcloud:
@claude if what your original diagnosis shows is true, sending a global ui message would have no impact whatsoever. You need to find the frame that would normally triage the packet for the button refresh, and sendframeuimessage

PR #2462, opened 2026-08-08
feat(LootBeacons): beacon drops by item name

3vcloud:
Requested in #gwtoolbox: expand the Loot Beacons settings with a list of string/colour pairs, checked against the decoded item name.

## What's new
- **Beacon by item name** section in the Loot Beacons settings: a list of rules, each with an on/off checkbox, a text box, a colour wheel and a delete button, plus an **Add item name** button.
- A rule matches when its text appears anywhere in the item's decoded name (case-insensitive), so `scroll`, `dye` or `glob of ectoplasm` all work.
- Name rules are checked **before** value and rarity, and the first matching rule in the list wins, so the list doubles as a priority order.
- Persisted as `name_beacons` in the module's settings JSON.

## Implementation notes
- Item names are decoded asynchronously via `GuiUtils::EncString` (`single_item_name`, falling back to `name_enc`) and cached per encoded string; an item simply doesn't match by name until its decode lands, a scan tick or two later.
- Rule patterns are lowercased once per 250ms scan, not per item.
- Docs page `site/src/content/docs/loot_beacons.mdx` updated; `npm run build` in `site/` passes.

Not built/tested in-game here (no MSVC toolchain in this environment) - worth a quick check that the new rows lay out sensibly in the settings panel.


henderkes:
@claude fix the build conflicts

henderkes:
@claude bitch please

henderkes:
@claude I told you to fix the build, it's still broken

PR #2630, opened 2026-09-16
Fix Objective Timer

### LINKED ISSUE #2610: `Objective Timer does all runs via JSON format from runs folder does not work correctly'
Image

Image

Whenever the runs start via the Objectives timer, etc. it automatically disappear from the list, even though the lists are in the runs folder. To be clearer, each run stays in the timer even after resigning or exiting the game. etc.

This will need to be fixed.

RiftLurker:
Previous implementation used incorrect rect bounds to determine if the objective set should be virtualized.
Using cursor.x moves the rect into the frame making it show up at all. Adding row_height allows for an objective set to be partially drawn.

Fixes #2610


3vcloud:
Thanks for taking the time

PR #2631, opened 2026-09-17
Add Trifecta and Feasts to Pcons menu

kennethhyman:
This MR adds the logic for popping empowering feasts and heroes' trifecta via the toolbox cons menu.


kennethhyman:
I made some assumptions around behavior that may not be valid, perhaps this calls for a setting that checks if people want to "use empowering feasts if any con is missing", and same for trifectas. Personally, I'd just rather that they stay at the top of the settings so that logic is evaluated first

kennethhyman:
I've verified this functionality works jumping into a zone with a mix of cons and pcons enabled, with the trifecta and feast sitting at the top.

3vcloud:
Didn't test very well then - just from reading the code I can see that it won't trigger if you already have a cupcake on but you want to trigger the other effects

kennethhyman:
@3vcloud - Yes im aware; In my opinion that would be my desired behavior, but perhaps its best to add an option for "use empowering if missing any cons" and same for the trifectas?

3vcloud:
Ok, and what happens if you have feast and cupcake enabled when entering a zone?

BearsMan:
It will automatically pop all of your pcons, including your cupcakes again.

kennethhyman:
When testing, it did not pop the extra cupcake however it does depend feast being on the top of my pcon list

kennethhyman:
Reason being, It seems to evaluate in order, and cupcakes do not pop if you're under the effect of cupcakes, which is the same effect regardless of coming from feasts or a cupcake directly

kennethhyman:
One fix I'll add is waiting for all party members to be alive and loaded in before using trifecta

3vcloud:
> Reason being, It seems to evaluate in order, and cupcakes do not pop if you're under the effect of cupcakes, which is the same effect regardless of coming from feasts or a cupcake directly

This cannot be true. Effects sent async by the server.

kennethhyman:
@3vcloud - thanks for the clarification; I will rework the logic and ask for a re-review this morning. I was operating off of observation, so i appreciate the insight

GWToolboxdll/Windows/Pcons.cpp
```diff
@@ -727,6 +733,100 @@ bool PconCons::CanUseByEffect() const
...
+    if (!auto_deselect_covered_pcons_on_enable) {
+        return;
+    }
+    using namespace GW::Constants;
+    static constexpr DWORD covered_items[] = {
+        ItemID::Warsupplies, ItemID::Apples, ItemID::Corns,
+        ItemID::Pies, ItemID::Eggs, ItemID::Cupcakes
+    };
+    for (Pcon* other : PconsWindow::Instance().pcons) {
+        auto* generic = dynamic_cast<PconGeneric*>(other);
```
kennethhyman, line 777: I'm not a CPP dev, so I'm not sure if this is a reasonable way to handle this, but this is how my experience with other typing systems work out

GWToolboxdll/Windows/Pcons.cpp
```diff
@@ -99,6 +102,9 @@ void Pcon::SetEnabled(const bool b)
     *enabled = b;
     ResetCounts();
     Refill(refill_if_below_threshold && IsEnabled() && PconsWindow::Instance().GetEnabled());
+    if (b) {
+        OnEnabled();
```
kennethhyman, line 106: This hook is necessary for disabling redundant cons/pcons when enabling trifecta / feasts. This seemed to be the best way to handle the issue of wasting pcons/cons depending on order of the buffs taking effect

3vcloud:
Superseded by #2633, which merged the commits through 7f59b1f9a48ce722b198b2488d724a7749f84f7b and includes the follow-up pcon effect-handling changes. Closing this PR to exclude its later commits.

kennethhyman:
Thanks for the help on this @3vcloud - Could you help me understand what I can do better next time to smooth out the PR process? And any other feedback about the ways i chose to implement things compared to the merged #2633

3vcloud:
> Thanks for the help on this @3vcloud - Could you help me understand what I can do better next time to smooth out the PR process? And any other feedback about the ways i chose to implement things compared to the merged #2633

Comes down to knowledge about how the game works - anyone who can use a keyboard and mouse can ask AI to make code these days, but it won't consider or know how the game behaves. In this case, skill effects won't apply straight away because the server needs to receive your item use packet and then later on it'll give you the skill. This is the case for everything in pretty much every game; your actions are not synchronous with the effects that they have, so you need to add arbitrary sleeps in that would be long enough for the server to handle and provide the effects for your ping.

Also read and understand what the existing code is doing - other pcons were checking for a single effect, but now we have pcons that check for multiple effects; its much less code to amend the base class to handle an array of effects than try and roll your own code on top it that. In the end, I think only around 30 extra lines were needed rather than your ~90 lines.

Finally, actually test the code as a player - despite what you've said I'm pretty certain this implementation would have caused duplicate uses of pcons, and potentially even worse for the trifecta; you could have potentially triggered trifecta, then immediately after that essence, grail and armor if you have all 4 active on map entry.

I usually end up heavily refactoring PR's that come in anyway because its easier than leaving code reviews and all that bollocks, so your original code was helpful.

kennethhyman:
Got it, thanks for the feedback and for getting this implementation across the line; I still think there may be some value in the toggling/ untoggling logic, but i can cut a separate MR for that if it would be helpful. Hoping to be able to contribute more and more effectively in the future

PR #2633, opened 2026-09-18
Add feast and trifecta pcons

3vcloud:
Includes PR #2631 only through 7f59b1f9a48ce722b198b2488d724a7749f84f7b; later PR commits are intentionally excluded.\n\nFixes the Feast and Trifecta checks to use an item when any constituent effect is absent or expiring. Records expected effects when a pcon is used and suppresses retries for one second while the game applies those effects.


GWToolboxdll/Windows/Pcons.cpp
```diff
@@ -293,6 +312,17 @@ void Pcon::AfterUsed(const bool used, const int qty)
...
 
+void Pcon::RecordEffectTrigger(const GW::Constants::SkillID skill_id)
+{
+    effect_triggered_at[skill_id] = TIMER_INIT();
+}
+
+bool Pcon::IsEffectTriggerPending(const GW::Constants::SkillID skill_id) const
+{
+    const auto found = effect_triggered_at.find(skill_id);
+    return found != effect_triggered_at.end() && TIMER_DIFF(found->second) < 1000;
```
RiftLurker, line 306: I believe this would have to respect the pcons delay or it's going to slow down pcon usage for low pcon delays

3vcloud: post ui callback to be added, which would clear the map entry

